PR #7, opened 2019-06-10
feat(sig): Platform SIG proposal

robzienert:


aglover:
Only one lead for this SIG?

robzienert:
Closing, as there does not seem to be interest from the community in helping lead.

plumpy:
I would be interested in being a co-lead here if you still think it's worthwhile. I like tools.

If no one else is interested, however, perhaps it's not worth the full SIG treatment.

robzienert:
I've reopened this PR and updated the scope of the SIG, and updated the leadership.

aglover:
+1 from me. @duftler @demobox @rmeshenberg @pstout what say you?

AbdulRahmanAlHamali:
This seems interesting. I'm interested in passive attending to keep up with the updates and the flow of things (and if I ever can contribute a valuable opinion I would be happy to)

demobox:
"as well as Spinnaker extensibility and plugins" seem like topic areas that go beyond what I have in mind when I think of "tools".

@robzienert could you elaborate a bit?

robzienert:
@demobox I was aiming to keep an explosion of SIGs to a minimum. There's a need for a group to sync on tooling, but also a need for extensibility. Extensibility cannot be done without a good tooling story in my mind, so I squished them. Perhaps a different SIG name would help alleviate your concern?

demobox:
> Perhaps a different SIG name would help alleviate your concern?

Do you have any suggestions? I may also have an inaccurate picture of what we mean by "extensibility" - I imagine things like a plugin model, potentially an SDK to go with it, that kind of thing. "Tools" I associate more with things like our build/test/release infra (i.e. developer-facing tooling, not *user-facing* tools such as `hal` or `spin`).

Those seem rather different to me? E.g. I wouldn't necessarily imagine that a group that discusses how and where to run our CI to necessarily be the same group that debates the best plugin model for Spinnaker.

robzienert:
What about... "Platform SIG"? I do agree that extensibility isn't really a "tools" thing and doesn't fit all too well. Since we don't really have a strong attendance of SIGs at this point, and most people who would be a part of this SIG are already in a ton of SIGs, I'm trying to be sensitive to further meeting explosion.

plumpy:
Andrew, I definitely see your point, I imagine the overlap between "people who want to talk about the build system" and "people who want to talk about plugins" might be somewhat small.

But to Rob's point, neither one is likely large enough to warrant its own SIG. I am happy and excited to talk about both, but even if most attendees are only there for one topic or the other, that feels okay to me?

demobox:
> What about... "Platform SIG"

Sounds good to me. Agree that we want to avoid extra SIGs unless it really makes sense; fine with starting this as a "Platform SIG" and considering splitting should that ever seem necessary.

robzienert:
Updated following all of the good feedback.

link108:
I do agree that the scope seems a bit broad, I'd prefer to see a platform SIG that handles extensibility and another sig to handle build/test/release. That being said, it looks like there is consensus right now for combining the two. 

I'd like to see some representation from Armory on the SIG leadership as we have a vested interest in the project, both build/test/release and extensibility/plugins. I'm happy to be that person or we can suggest someone else.

robzienert:
Added @link108 as another SIG Lead.

sig-index.md
```diff
@@ -6,6 +6,7 @@ Don't see a SIG here that fits what you're looking for? [Submit a Proposal](sig-
 * [AWS](sig-aws/README.md)
 * [Spinnaker-as-Code](sig-spinnaker-as-code/README.md)
```
pstout: nit: Having sorted your addition into the correct place, would you please move this SIG down in the list.

duftler:
+1

PR #12, opened 2019-07-12
docs(roadmap): Initial high-level roadmap

robzienert:
I'd like the roadmap in the community repo to be theme-oriented. For tracking actual deliverables, we should adopt using Projects in `spinnaker/spinnaker`, with a project for each minor OSS release.

This will become a living document. SIGs and other organizations that are invested in contributing to major roadmap themes should open PRs to update this as necessary.

Currently some of these high-level themes are "led" by companies; which I don't particularly like. I'd like to replace these with working groups so that it's more inviting for community members to get involved in various efforts.

fyi @spinnaker/sc


christopherthielen:
love it

PR #13, opened 2019-07-12
feat(issues): New cloud provider proposal template

robzienert:


robzienert:
@spinnaker/sc fyi

ezimanyi:
> Do we feel answering these questions for some of the existing providers would be helpful?

Yeah, I think that would definitely be a good exercise.  I think at one point we had discussed trying to get SIGs for all of the providers, and this could be a good opportunity to do that (along with the other requirements).

ajordens:
> > Do we feel answering these questions for some of the existing providers would be helpful?
> 
> Yeah, I think that would definitely be a good exercise. I think at one point we had discussed trying to get SIGs for all of the providers, and this could be a good opportunity to do that (along with the other requirements).

I worry about SIG fatigue and question whether a sig for every provider is needed (or even feasible). 

I could get behind a general purpose provider sig given there is a reasonable amount of commonality across provider implementations. 

This is independent of clearly identified owners for each provider, that’s still required.

robzienert:
> I worry about SIG fatigue and wonder whether a sig for every provider is needed (or even feasible).

If there's not enough interest in a cloud provider integration to warrant a SIG - should it even be a part of core Spinnaker? My opinion is a definite "no". I don't want to have bespoke methods of staying in contact with various cloud provider maintainers. A SIG provides a paved path to issue triage, bubbling up of issues to the TOC or even SC, and a place for community members who are not owning maintainers to have their voices heard (via the SIG regular meetings).

I do appreciate thinking about SIG fatigue, but I don't think cloud providers is a place where we want to go lean. I do like the idea of a general cloud provider SIG, though, but that's going even further down the path of SIG explosion. 😆  Integrations like CI systems that would land in Igor? Yeah, I think we could have a single, general purpose SIG there.

ajordens:
Much of the above makes sense for a provider in Spinnaker core. 

Do we feel there’s a difference between incubating a new cloud provider and it being formally recognized as part of core. 

I don’t believe there’s even a clear description of the pieces relevant to a full blown provider implementation, it’d be nice to have that available as part of a checklist as it generally comes up in Slack. Should we prioritize this with the TOC?

Nuanced but I’d be satisfied if the SIG expectations only came into play once a provider left incubation.

Such a general purpose sig could potentially be oriented around helping new providers get off the ground and shepherding them through the implementation. Just an idea, not entirely convinced of it over just using  #dev.

robzienert:
Until we get to a place where we can have a true "Core Spinnaker", built and distributed independently of its cloud providers, I think we have to consider any new cloud provider part of core Spinnaker?

We should definitely amend the Cloud Provider requirements doc to outline what all is involved in making a full-blown cloud provider, because it is non-trivial and I suspect organizations don't know what all is involved until they're halfway in. I'm going to merge this for the time being, but I definitely think this merits additional conversation. 👍

PR #22, opened 2019-08-21
Proposed updates to governance policy from SC meeting

aglover:
First pass at refining aspects of the policy. 
1. clarified TOC nomination and approval process
2. same for SC
3. added notion of term limits + consecutiveness 
4. Chair person for both TOC and SC
5. Calendar year on Spinnaker Summit

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -75,6 +75,8 @@ Qualifications
   * Terms should be staggered within the TOC and off by 6 months from SC 
   * Staggering should result in approximately half of the committee up for election each year
   * Term starts mid-calendar year (i.e. July 1st) 
+  
+Election to the TOC requires a majority of the SC. 
 
 Logistics
 
@@ -119,7 +121,7 @@ Current SC members:
 * Ruslan Meshenberg (Netflix)
 * Andrew Philips (Google)
 
-Election to the SC requires a majority of the SC voting. If a person is up for reelection they are not allowed to vote for their reelection. 
+Election to the SC requires a majority of the SC. If a person is up for reelection they are not allowed to vote for their reelection. 
 
 ## Special Interest Groups
 
--- a/governance.md
+++ b/governance.md
@@ -86,7 +86,7 @@ Current TOC members:
 * Rob Zienert (Netflix)
 * Travis Tomsu (Google)
 
-The TOC should remain with an odd number of members (between 3 to 7) so as to avoid deadlock. In the event that someone from TOC leaves prematurely, the SC should be proactive in filling the spot to ensure an odd number of members. The TOC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.  
+The TOC will have three, five, or seven members, as deemed appropriate by the Steering Committee. In the event that someone from the TOC leaves during their term, the Steering Committee shall appoint someone to fill the remainder of the term. Each year, the TOC will nominate a member to serve as the Chairperson for that year. The Chair's duties will include calling and runnings meetings, calling votes, and ensuring that meeting notes are recorded. 
 
 ### Steering Committee
 
@@ -118,7 +118,7 @@ Current SC members:
 * Ruslan Meshenberg (Netflix)
 * Steven Kim (Google)
 
-The SC should remain with an odd number of members so as to avoid deadlock. The SC will nominate a chairperson for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes. 
+The Steering Committee will initially have five members. The Steering Committee may adjust the size of the committee each year as they deems appropriate, with a bias towards an odd number of members to avoid voting deadlocks. In the event that someone from the Steering Committee leaves during their term, the Steering Committee shall appoint someone to fill the remainder of the term. Each year, the Steering Committee will nominate a member to serve as the Chairperson for that year. The Chair's duties will include calling and runnings meetings, calling votes, and ensuring that meeting notes are recorded. 
 
 ## Special Interest Groups
 
--- a/governance.md
+++ b/governance.md
@@ -86,7 +86,7 @@ Current TOC members:
 * Rob Zienert (Netflix)
 * Travis Tomsu (Google)
 
-The TOC should remain with an odd number of members so as to avoid deadlock. The TOC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.  
+The TOC should remain with an odd number of members (between 3 to 7) so as to avoid deadlock. In the event that someone from TOC leaves prematurely, the SC should be proactive in filling the spot to ensure an odd number of members. The TOC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.  
 
 ### Steering Committee
 
--- a/governance.md
+++ b/governance.md
@@ -118,7 +118,7 @@ Current SC members:
 * Ruslan Meshenberg (Netflix)
 * Steven Kim (Google)
 
-The SC should remain with an odd number of members so as to avoid deadlock. The SC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.  
+The SC should remain with an odd number of members so as to avoid deadlock. The SC will nominate a chairperson for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes. 
 
 ## Special Interest Groups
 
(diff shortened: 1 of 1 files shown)


governance.md
```diff
@@ -70,8 +70,10 @@ Duties
 
 Qualifications
 
-* Nominated by community
-* Appointed by Steering Committee
+* Nominated and Appointed by Steering Committee
+  * 2 year terms, term limit of consecutive 4 years (after, individual must take off 1 year)
```
demobox:
```suggestion
  * 2 year terms, term limit of consecutive 4 years (after, individual must take 1 year off)
```

demobox: Also, for clarification, does this mean that after a 1yr break, a user can come back and serve for another consecutive 4 years?

governance.md
```diff
@@ -70,8 +70,10 @@ Duties
...
+  * Terms should be staggered within the TOC and off by 6 months from SC
+  * Calendar year starts with annual Spinnaker Summit
```
demobox: What if this were to move quite dramatically, e.g. to the spring?

governance.md
```diff
@@ -84,7 +86,7 @@ Current TOC members:
 * Rob Zienert (Netflix)
 * Travis Tomsu (Google)
 
-The technical oversight committee is initially seeded with core members from Netflix and Google. We will leave further details such as number of seats, qualifications, tenure, TBD for the time being.
+The TOC should remain with an odd number of members so as to avoid deadlock. The TOC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.
```
demobox: Thoughts on limits? E.g. >1? Any ceiling?

governance.md
```diff
@@ -96,6 +98,13 @@ Duties:
 * Address questions or concerns from the community about project culture, structure, and clarifications of roles, responsibilities and duties
 * Ratify new SIGs
 
+Qualifications:
+
+* Nominated and Appointed by Steering Committee
```
demobox: See comments above. If an SC member is up for re-election, do they get to take part in the nomination and/or vote?

governance.md
```diff
@@ -109,7 +118,7 @@ Current SC members:
 * Ruslan Meshenberg (Netflix)
 * Steven Kim (Google)
 
-The steering committee is initially seeded by Netflix and Google as the project matures and its structure stabilizes. We will leave further details such as number of seats, qualifications, tenure, appointing measures, TBD for the time being.
+The SC should remain with an odd number of members so as to avoid deadlock. The SC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.
```
demobox: See comments above about limits

governance.md
aglover: Yes, I think so.

governance.md
aglover: Fair enough - we could just make it easy w/calendar year. You ok w/that?

governance.md
aglover: YES! Great call. We should define a ceiling - anywhere between 3 and 7 seems reasonable to me. You?

governance.md
aglover: I guess we should add language here - my thought is that a member cannot vote for their own re-election.

governance.md
demobox: 3 to 7 also sounds good to start with. We could always increase the upper limit if needed via another vote

robzienert: 3 to 7, excluding even numbers. In the event that someone from TOC leaves prematurely, the SC should be proactive in filling the spot so we're back to an odd number of members. Thoughts?

ajordens: Sounds good to me.

governance.md
ajordens: Does that create the potential for a split vote? (chair person breaks a tie?)

demobox: > Does that create the potential for a split vote?

Er, um, indeed. Hrm. Chairperson breaking the tie sounds good to me

governance.md
aglover: Adding this language.

governance.md
aglover: I think I misread the initial question. If an SC member is up for re-election, they have full voting rights until they are removed. That means we need to prepare for elections before the end of the year. The person up for re-election shouldn't be able to vote for themselves though. And then in that case, the chairperson breaks the tie. Thoughts?

demobox: > The person up for re-election shouldn't be able to vote for themselves though. And then in that case, the chairperson breaks the tie. Thoughts?

That was my reading of @ajordens' suggestion - +1 to that from me

governance.md
```diff
@@ -59,7 +59,7 @@ If you’d like to nominate yourself as an Approver, please create an [Issue in
 
 ### Technical Oversight Committee
 
-The TOC is responsible for the overall technical management of the project, ultimately managing the day-to-day running of the project.
+The TOC is responsible for the overall technical management of the project, ultimately managing the day-to-day running of the project.
```
demobox:
```suggestion
The TOC is responsible for the overall technical management of the project, ultimately managing the day-to-day running of the project.
```

governance.md
```diff
@@ -84,7 +86,7 @@ Current TOC members:
 * Rob Zienert (Netflix)
 * Travis Tomsu (Google)
 
-The technical oversight committee is initially seeded with core members from Netflix and Google. We will leave further details such as number of seats, qualifications, tenure, TBD for the time being.
+The TOC should remain with an odd number of members (between 3 to 7) so as to avoid deadlock. In the event that someone from TOC leaves prematurely, the SC should be proactive in filling the spot to ensure an odd number of members. The TOC will nominate a chair person for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.
```
demobox: [minor] To avoid potential ambiguity about whether "between" does or does not include the boundaries, perhaps phrase simply as "...should remain with 3, 5 or 7 members so..."?

governance.md
```diff
@@ -109,7 +118,7 @@ Current SC members:
 * Ruslan Meshenberg (Netflix)
 * Steven Kim (Google)
 
-The steering committee is initially seeded by Netflix and Google as the project matures and its structure stabilizes. We will leave further details such as number of seats, qualifications, tenure, appointing measures, TBD for the time being.
+The SC should remain with an odd number of members so as to avoid deadlock. The SC will nominate a chairperson for a 1 year term. Their duty is to run/call meetings, take meeting notes, and call votes.
```
demobox: Any upper or lower limits desired here? I'd say at least 3?

governance.md
pstout: I would suggest replacing the new text with the following:

The TOC will have three, five, or seven members, as deemed appropriate by the Steering Committee.  In the event that someone from the TOC leaves during their term, the Steering Committee shall appoint someone to fill the remainder of the term.  Each year, the TOC will nominate a member to serve as the Chairperson for that year.  The Chair's duties will include calling and runnings meetings, calling votes, and ensuring that meeting notes are recorded.

governance.md
pstout: What happens if the chairperson is the one who is up for reelection and there is a tie?  I would propose instead that we simply state that a majority of the SC is needed to be elected (or reelected), which means that a person is not elected in the case of a tie.  If the committee is that split about a candidate, then I worry about their being a cohesive part of the leadership.

governance.md
pstout: Given that the Steering Committee has the ability to rewrite these guidelines, I am how much restrictive prose to write here.  I would suggest replacing the new text with the following:

The Steering Committee will initially have five members.  The Steering Committee may adjust the size of the committee each year as they deems appropriate, with a bias towards an odd number of members to avoid voting deadlocks.  In the event that someone from the Steering Committee leaves during their term, the Steering Committee shall appoint someone to fill the remainder of the term.  Each year, the Steering Committee will nominate a member to serve as the Chairperson for that year.  The Chair's duties will include calling and runnings meetings, calling votes, and ensuring that meeting notes are recorded.

duftler: I like the way this reads. I also agree that writing up a bunch of additional constraints doesn't seem worthwhile since as you pointed out, the SC can just change the policy anyway.

aglover:
Went through the comments again and tried to incorporate the feedback into the proposed changed document. Please take a look @demobox @robzienert @pstout @duftler @rmeshenberg @ajordens and either make new suggestions or vote +1 or -1. 

Thank you!

demobox:
> Please take a look demobox 

Suggested a minor correction, otherwise looks good to me

governance.md
pstout: @aglover I don't see this thread reflected in the changes to this file.  Is that intentional?

pstout:
+1

duftler:
+1

PR #23, opened 2019-08-22
Replace Steven Kim with Andrew Phillips on SC.

duftler:
Propagating change from https://github.com/spinnaker/community/pull/21.


duftler:
fyi @aglover

PR #40, opened 2019-09-23
feat(docs): Adding RFC template

robzienert:
Aiming to have some better consistency and depth across our proposals.


plumpy:
Yeah, this is great, thanks!

PR #41, opened 2019-09-27
feat(rfc): Kubernetes operator for Spinnaker

ncknt:


ncknt:
@ezimanyi @ethanfrogers

rfc/kubernetes-operator.md
```diff
@@ -0,0 +1,89 @@
...
+- How should Halyard itself be managed? Locally spun up at CI time or as a long-time running service?
+- How do you template the configuration if you have multiple running instances?
+- How do you pass credentials to Halyard, like kubeconfig files or AWS credentials?
+- Users are left with a lot of activities that Halyard doesn’t handle such as exposing Spinnaker and cleaning up old secrets.
+
+Finally, there’s the impression that Spinnaker is not “Kubernetes native” because it cannot be installed and managed with the usual administrator toolchain (kubectl). Talking to Spinnaker administrators, we’ve realized this is a pretty widely shared feeling. They’re already proficient at using kubectl and are reluctant to learn and configure yet another new tool.
+
+## Timeline
+
+Current MVP can deploy and expose Spinnaker. Preflight checks will be added in the coming weeks. Moving the operator to Spinnaker's GitHub account involves publishing new images to gcr and cleaning the build.
```
ezimanyi: What sort of preflight checks will be added (both the short-term ones and the longer-term ones) and how will they be implemented?

The implementation of preflight checks in Halyard has a number of issues, and I'd like to understand how this implementation will be different (and ensure we won't just repeat/reuse the Halyard implementation).  A couple of the issues in Halyard:
* The checks are performed using different credentials than Spinnaker will use when it's actually running
* In order to implement the checks, Halyard needs a jar-level dependency on several of the other microservices.  These microservices don't expect to be consumed as a library like this, and so very frequently break Halyard.

rfc/kubernetes-operator.md
```diff
@@ -0,0 +1,89 @@
...
+
+It's another project. The operator is also a new deployment running in the user's cluster. If the operator is strictly a tool to configure (vs monitor), that may seem overkill.
+
+## Prior Art and Alternatives
+
+We considered that [operator](https://operatorhub.io/operator/spinnaker-operator) but couldn't find good documentation, the code, or users using it. It also seems to be running a Halyard job in the cluster. Feel free to correct me.
+
+## Known Unknowns
+
+The `spec` of the CRD is still in flux. As we add more features, it will need to evolve before being stable. As an example, new fields are needed to track validation options. The existing `status` is also a first stab.
```
ezimanyi: Can this link out to the spec of the CRD?  Will the CRD need to be updated as we add config options to the hal config, or does it just reference that as an opaque unit?

rfc/kubernetes-operator.md
```diff
@@ -0,0 +1,89 @@
...
+
+There's currently no telemetry baked in.
+
+## Risks
+
+The implementation relies on Halyard's deployer being able to generate manifests instead of applying them.
+
+## Future Possibilities
+
+- More preflight and scheduled checks for connectivity, authorization: we'd for instance like to move some of the cloud provider account check to preflight check to speed up Clouddriver startup. The health of Spinnaker could be visible via `kubectl describe` and events.
```
ezimanyi: As the operator performs more and more preflight checks, it will need to understand more and more about the implementation details of each microservice.  How do you see this being implemented in a way that still maintains separation of concerns?

Will the operator need to know how to talk to a Jenkins server so that it can preflight things before spinning up Igor?  (As one example.)

rfc/kubernetes-operator.md
```diff
@@ -0,0 +1,89 @@
...
+| **Status**     | **Proposed** |
+| **RFC #**      | https://github.com/spinnaker/community/pull/41 |
+| **Author(s)**  | Nicolas Cohen (`@ncknt`) |
+| **SIG / WG**   | Kubernetes SIG |
+
+## Overview
+
+Halyard is great at offering a simple way to configure Spinnaker services but is not easy to operationalize. We've come up with an MVP of a Kubernetes operator for Spinnaker that allows administrators to manage Spinnaker with kubectl (or whatever Kubernetes tool).
+
+We're proposing to move the operator to Spinnaker's GitHub: https://docs.google.com/document/d/1Fs-xK7sOZ4YLSw1ae_cQ_rsB0kMNon9AfYli2a5Udhs/edit?usp=sharing
```
ezimanyi: What are the success criteria for this new project, and plans for ongoing maintenance?  Will this initially be labeled 'alpha' and have a plan to deprecate if it does not gain traction, or are we effectively promising ongoing support for this from day 1?  Is there a SIG or working group that will own the maintenance of this?  (I see the Kubernetes SIG is listed above, but I think there should be a more precise and specific group responsible for maintenance of this at least for the near future.)

rfc/kubernetes-operator.md
```diff
@@ -0,0 +1,89 @@
...
+
+Current MVP can deploy and expose Spinnaker. Preflight checks will be added in the coming weeks. Moving the operator to Spinnaker's GitHub account involves publishing new images to gcr and cleaning the build.
+
+## Design
+
+See https://docs.google.com/document/d/1Fs-xK7sOZ4YLSw1ae_cQ_rsB0kMNon9AfYli2a5Udhs/edit#heading=h.7vexvfs9sl5e
+
+### Dependencies
+
+No new dependencies to existing projects. Some (relatively small) changes are necessary in Halyard to:
```
ezimanyi: I noted this on the design doc, but adding here as well---I'd like to see a more shared ownership of Halyard if we're going to be building additional dependencies on top of Halyard.  I review a large percentage of the PRs to Haylard and would like to see that responsibility shared more; I think that if this will be built on top of Hayard it makes sense for the maintainers of this operator to take a larger role in maintaining Haylard.

rfc/kubernetes-operator.md
ncknt: Added the link (https://github.com/armory/spinnaker-operator/blob/master/deploy/crds/spinnaker_v1alpha1_spinnakerservice_crd.yaml). The CRD does not need to be changed with different versions of Halyard and could potentially work with all recent versions of Halyard (well minus the fact they need to have the manifest endpoint). We just need to edit when we add features to the operator itself (e.g. control how to do validation, how to expose Spinnaker services, etc.)

The Halyard side of things is just a `configMap` (or a `secret` if there's demand) referenced from the `SpinnakerService`.

ncknt: To add to it, it's possible we start pulling some of the fields from the config to the CRD itself to enable a better `kustomize` experience. For instance, `version` is an obvious candidate.

rfc/kubernetes-operator.md
ncknt: We have a long list of common issues. Most of the time, it's about testing connectivity and auth. In no particular order:
- Kubernetes accounts: check connectivity, namespace, kinds, CRDs like Clouddriver does on startup. When enabled, we'd like to turn it off for CD so it can boot up faster on all pods. The check can also be done on a regular basis.
- Other provider accounts: connectivity and authorization, IAM role properly set up etc.
- Other type of checks: Version, single Spinnaker per namespace, etc.

We're working on a formalizing the framework. Happy to share that soon.

rfc/kubernetes-operator.md
ncknt: That's the idea. I realize it's no small feat and getting 100% preflight check will take a long time but we believe we can achieve address the most common issues reported relatively fast.

rfc/kubernetes-operator.md
ncknt: I'm not sure how to measure success yet but we're really after installing Spinnaker on Kubernetes faster, will less issues. We want to appeal to users who are after a Kubernetes native tool and feel this is a good first step. We'll track usage and success rate for our customers but there's no telemetry baked in.

It's currently alpha and the CRD is subject to change. We do plan to support and improve the operator (we have engineers actively working on it) and hopefully get feedback and help from the community. We created a `kubernetes-operator` Slack channel and are open to making a new SIG if the Kubernetes SIG is not the right group.

ethanfrogers: I don't see why we couldn't form a WG under the Kubernetes SIG for this project specifically.

rfc/kubernetes-operator.md
ezimanyi: Thanks for the details---I'd definitely like to understand the technical approach to how these pre-flight checks will be performed before signing off on this proposal.  I definitely don't want to create a situation where we create hard-to-maintain strong dependencies between the operator and implementation details of the microservices, so am very interested in the approach here.

rfc/kubernetes-operator.md
ezimanyi: Yeah, I think a WG under the Kubernetes SIG would be a good solution; mostly I want to make sure there's a direct owner of this for the forseeable future and maintenance of this doesn't immediately fall to the Kubernetes SIG which already has a fair amount of work on its plate.

ncknt: Another possibility is a SIG install to deal with all installation matters. Regardless of how we organize it, I'm happy to take ownership of it.

rfc/kubernetes-operator.md
ncknt: That's a fair point. Thanks for putting it up with all the work. We're ready to step up and help with the load on Halyard.

ezimanyi: Thanks, that would be great!

ajordens:
Are we content to just leave this proposal sitting?

ethanfrogers:
This kind of fell off my radar. I'm fine with the proposal as is but I know @ezimanyi and @ncknt have been chatting so I'll defer to them on where we stand currently. If there have been any updates, I'd like to see them reflected in this doc, however.

ezimanyi:
My primary concern initially was that this was building on top of Halyard, which is already pretty complex and didn't have many people owning/maintaining it other than me. This has been addressed as others have started reviewing more Halyard PRs, and I've also been discussing with @ncknt how we can build a much smaller, simpler Halyard that meets the requirements of the operator.  A separate proposal will follow for that, but in the meantime I think this proposal here is reasonable.

ncknt:
Thanks @ezimanyi. We've been working on a new version of the CRD without the `configMap` and have had good success with it so far. A better overview of the operator can be found in the readme of the repo and linked docs: https://github.com/armory/spinnaker-operator

We'll be submitting more changes to Halyard and as Eric mentioned work on the longer term vision for Halyard separately.

PR #42, opened 2019-10-08
docs(governance): Adding SIG Leads as community role

robzienert:


governance.md
```diff
@@ -57,6 +57,22 @@ Qualification guidelines
...
+Duties
+
+* Set agenda for, facilitate, and drive SIG meetings
+* Review and sponsor proposals related to the SIG
+* Triage inbound spinnaker/spinnaker issues related to the SIG
+* Review and merge or decline pull requests from the community
+
+Qualifications
+
+* Approver status
```
ethanfrogers: What happens to current SIG leads who aren't currently approvers?

robzienert: Good question, my thoughts is that nothing changes who is holding positions today. I'm just trying to clarify responsibilities here, not shake things up.

duftler:
I am in favor of adding this to the governance documentation.

+1

governance.md
```diff
@@ -57,6 +57,22 @@ Qualification guidelines
 
 If you’d like to nominate yourself as an Approver, please create an [Issue in the community repository](https://github.com/spinnaker/community/issues). 
 
+### SIG Leads
```
demobox: How is this intended to relate to the "SIGS" section later in the doc? https://www.spinnaker.io/community/governance/#sigs

It states there that "SIGs and their leads are ratified and greenlit by the steering committee", which isn't in conflict with the qualifications described here, but adds an additional requirement (ratification by the SC) that isn't mentioned here.

The [SIGS page](https://www.spinnaker.io/community/governance/sigs/) also mentions that "Each SIG has one or more “leads”, who own the SIG’s agenda", which, while again not in conflict, is a different description of the duties of a SIG lead as described here.

Thoughts on one or either of:

* updating the description of SIGS later in this doc to refer to this new role
* updating the SIGs page to link to the new role description

?

To also take a step back: is there a particular issue/set of issues we're looking to address by formalizing the role in this way?

governance.md
pstout: Are there any who are not approvers?  If so, why do we think that this should be a requirement?  If we agree that it should be a requirement, what are the non-approver SIG leads going to do to become approvers?

All that said, I am not clear I agree that it should be a requirement for all SIGs.  For example, what value does it bring to the community to require that the leads for the Summit SIG need to be approvers?  Maybe I am missing something about the scope of what being an "approver" is.

governance.md
pstout: I actually think there is a conflict between line 74 below, "Self-nominated, elected by SIG attendees", and the line in the "SIGS" section of governance document, "SIGs and their leads are ratified and greenlit by the steering committee".

I like both of Andrew's suggestions and I am also interested in the answer to his question.

governance.md
plumpy: The Docs SIG is a great example of one where the leads are not approvers (except of the docs repo) and probably aren't likely to ever become full approvers, but nor does that seem inappropriate, IMO.

ethanfrogers: Security SIG is another example, as well.

robzienert: Very well, I'll just remove this from the qualifications. Definitely agree that Summit / Docs SIGs shouldn't need approvers. Security SIG, however, seems not like the others.

plumpy: "For SIGs that are primarily about technical issues, the SIG leads should be approvers"? That's still wishy-washy, so maybe that's just making it too complicated.

governance.md
robzienert: SIG Lead nomination: 

I'm happy to fall in-line with the rest of the docs so that Leads are always ratified by the SC. As we get more SIGs and the community grows, I fear gating all SIG Lead changes by the SC could stall momentum. 

By making SIG Lead nomination and election community-driven, a SIG could be a little more autonomous than they are today. We'd still likely want more language around how many Leads must be maintained by a SIG and around diversity of represented invested organizations (such that a company can't steamroll leadership of a SIG).

Sidenote: If ratification is required for SIG Lead changes, we'll also need to explore term limits and replacement processes, which it looks like we're missing today. 

> To also take a step back: is there a particular issue/set of issues we're looking to address by formalizing the role in this way?

This is in response to direct feedback by the SIG Leads to the TOC in that they don't have a clear understanding of what they are supposed to be doing for the community or their SIG(s), as well as ghosting / too-light-touch SIG Leads. SIG Leads that have been around since the beginning understand, but only due to tribal knowledge.

There has been some impression that SIG Leads would only be responsible for setting up and running the SIG meetings, when in fact there's a lot more project and product management tasks that they need to perform to have a SIG be effective. Prior to this, a lot of day-to-day responsibility was implicitly falling on the TOC that doesn't belong.

robzienert: Taking a step back, perhaps SC ratification of SIG Leads is best for the short-/medium-term while SIG attendance is low. 🤔

duftler: I also think it's ok to leave ratification with the SC for now. Maybe we can add a standing SC meeting agenda item to ensure we work through any vote-ready issues in each meeting.

governance.md
link108: We may want to include "Belongs to an organization with significant stake in project" as a qualification to being a SIG lead.

governance.md
demobox: > This is in response to direct feedback by the SIG Leads to the TOC in that they don't have a clear understanding of what they are supposed to be doing for the community or their SIG(s), as well as ghosting / too-light-touch SIG Leads

Ack, thanks for the additional details!

robzienert:
Folded in all the feedback. Take another look, I think this is vote-ready at this point?

duftler:
+1 (in its final form)

governance.md
```diff
@@ -57,6 +57,18 @@ Qualification guidelines
...
+
+SIG Leads are responsible for management of a particular Special Interest Group within the project, managing day-to-day operations of their interest scope.
+
+Duties
+
+* Set agenda for, facilitate, and drive SIG meetings
+* Review and sponsor proposals related to the SIG
+* Triage inbound spinnaker/spinnaker issues related to the SIG
+* Review and merge or decline pull requests from the community
```
pstout: Excuse my ignorance of GitHub workflow, but are Leads who are not Approvers going to be able to do this?

robzienert: No, they won't, which is why having a qualification of being an Approver was part of the original PR. The issue is that there are SIGs that do not have anything to do with the actual code (community, documentation, etc). Perhaps there should be language that Approver is required for Leads who are in SIGs involving the codebase? I could use help with this.

aglover: I've updated things -- I'm going to merge this PR and we can refine the language in a subsequent edit of the document.

PR #43, opened 2019-10-09
feat(rfc): Plugins

robzienert:
⚠️ This proposal is not intended to be all-things-considered. We're aiming to have a general direction down and then learn and adapt as development goes.

---
This change is [Reviewable](https://reviewable.io/reviews/spinnaker/community/43)

### DIFF
--- a/rfc/plugins.md
+++ b/rfc/plugins.md
@@ -1,0 +1,343 @@
+# Plugins Proposal
+
+| | |
+|-|-|
+| **Status** | **Proposed**, Accepted, Implemented, Obsolete |
+| **RFC  Num** | [#43](https://github.com/spinnaker/community/pull/43) |
+| **Author(s)** | Rob Zienert (`@robzienert`), Cameron Motevasselani (`@link108`) |
+| **SIG / WG** | Platform SIG |
+| **Obsoletes** | [spinnaker/spinnaker#4181](https://github.com/spinnaker/spinnaker/issues/4181) |
+
+## Overview
+This proposal builds atop prior proposals and proof of concepts for a plugin system within Spinnaker. 
+This proposal is for an experimental body of work spanning kork and multiple services including Deck, and eventually, standardization of some development lifecycles of Spinnaker itself.
+
+### Goals and Non-Goals
+* Goals
+	* Support plugins on both the frontend & backend services
+	* Support plugins spanning multiple services
+	* Support in-process and RPC-backed plugins for backend services
+	* Identify up/downgrade processes for plugins
+	* Provide tooling for developers and operators
+* Non-goals
+	* Support for alternative runtimes (e.g. Lambda, Docker; to be addressed later)
+	* Identification of all extension points within Spinnaker
+	* Event-based plugins
+	* Finalization of extension point acceptance processes
+	* Finalization of plugin development and publishing workflow
+	* Finalization of plugin operator workflow
+
+It should be clear from these lists that this proposal is focusing on the service-side of plugins and to be a launching point for discovery and subsequent narrowly-focused proposals of ecosystem work.
+
+Initial plugin targets will be:
+
+* Pipeline Stages
+* Pipeline SpEL functions
+* Micrometer backends
+* Gate proxy endpoint configuration
+
+## Motivation and Rationale
+Extending Spinnaker today falls into two categories, 1) writing Java code that directly extends internal APIs, or 2) usage of web hook stages or preconfigured jobs.
+
+In the first case, a native Spring extension model, extensions need to be written on the JVM and released in lock-step with service releases. 
+Further, it usually requires running the Spinnaker ecosystem to validate correct behavior - a technical hurdle that many possible contributors do not have the time or patience for. 
+In the second case, extensions can be written and tested outside of the Spinnaker release cycle, but can only affect functionality of Spinnaker in the narrow scope of a pipeline stage.
+
+In an effort to lower the bar for Spinnaker development, a native plugin model is being proposed that can be used for federated development of Spinnaker. 
+This native plugin system will need to support clearly defined extension points within Spinnaker at both service- and product-level abstractions, from either an in-process or remote runtime model.
+
+By creating this plugin model, Spinnaker will be able to iterate functionality faster as changes will be more easily developed, distributed and installed without adding bloat to the core product offering. 
+This work aims to be a quality of life improvement for Spinnaker developers, operators, and users alike.
+
+Early stakeholders for this effort are:
+
+* Netflix Delivery Engineering
+* Netflix Cloud Database Engineering
+* Armory
+
+We have explicitly not called out the Spinnaker OSS community, as the early deliverables of this effort are experimental and won’t be subject to backwards compatibility constraints.
+Community members are welcome to be involved early nonetheless, provided acceptance of instability.
+
+## Timeline
+Proof of concepts were developed both by Armory and Netflix independently in 2019Q3. 
+In Q4, we plan to deliver an alpha-quality plugin implementation that exhibits capability of extending multiple services as well as Deck, and include a basic distribution story for both in-process and out-of-process extensions.
+
+Development will continue into and through 2020. 
+If the plugin proposal is successful, plugin development and thereby the core plugin framework will continue iteration for the foreseeable future.
+
+## Design
+### Taxonomies
+| Term | Definition |
+| - | - |
+| **Extension Point** | A strongly typed service- or library-defined point of extension |
+| **Plugin** | An implementation of an Extension Point |
+| **Installed Plugin** | A plugin known to Spinnaker, regardless of enabled state |
+| **Enabled Plugin** | A plugin that is installed and actively participating in the application lifecycle |
+| **Disabled Plugin** | A plugin that is installed, but not participating in the application lifecycle |
+
+### Multiple Plugin Runtimes
+There’s a place and a time for different plugin runtimes. 
+In-process extensibility offers a lot of power and efficiency at the cost of higher coupling to core service rollouts, whereas a remote plugin runtime trades runtime efficiency and reliability for development, release independence, and choice of technologies (language, frameworks).
+
+The goal of this proposal is to offer the best of both worlds by allowing extension point developers (Core Spinnaker developers) to decide what runtime(s) are the best for the extension.
+
+Extension points will be defined and exposed via PF4J.
+
+### PF4J as a Foundation
+[PF4J](https://github.com/pf4j/pf4j) is to be used as the foundation for all plugin functionality. 
+While designed for in-process extensibility, we’ll be using PF4J to define both the contract for in-process plugins, as well as remote plugins. 
+Alternative (co-process, remote) runtimes would also be available but will be covered later.
+
+PF4J has a lot to offer out of the box that we would need to build ourselves, regardless of final implementation. 
+It doesn’t go as far as OSGI, but provides basic (yet extensible) ClassLoader isolation, a simple extension contract that we can use broadly, and a standardized manifest format. 
+PF4J was originally built for creating in-process plugins only, but its own extension points make it feasible to enable remote plugins using the same contract. 
+What runtime a particular plugin uses would be determined by a combination of: 1) The extension point developer allowing remote invocation, and 2) the plugin developer writing the plugin for remote invocation.
+
+In the case of a remote plugin, the plugin developer will have the freedom to choose language, framework, and runtime. 
+For any remote plugin, an in-process counterpart PF4J plugin will be required to enable communication between a service and the plugin. 
+For example, a remote plugin that exposes its integration through a gRPC server would also need a gRPC transport plugin on the service so that communication between the two can be configured correctly. 
+Other transport plugins, such as Docker co-process, Functions, HTTP, and so-on could also be developed.
+
+An extension point looks like the following:
+
+```kotlin
+// Definition of the PluginStage extension point.
+// The annotation defined here also informs the Spinnaker
+// plugin framwework which runtimes are acceptable. This
+// annotation would additionally be used for other metadata
+// Spinnaker needs to know atop whatever PF4J offers out of
+// the box.
+@SpinnakerExtension(allowRemote = true)
+class PipelineStageExtension : ExtensionPoint {
+  // ...
+}
+```
+
+It is highly recommended to read the documentation of PF4J to fully understand the scope of this proposal: This proposal will not go in depth on PF4J internals or taxonomy.
+
+### Front50 as the Source of Truth
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. 
+Whether plugins are implemented as in-process or not, Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled.
+
+The Front50 plugin registry will also be exposed through Gate, and can be consumed by Deck to load the enabled plugins with UI components on application startup.
+
+Plugins can also be enabled or disabled at runtime without redeploying a service. 
+This flag is sourced from Front50, which all services will be listening for change events from. Whether this change event is published or polled for initially is unknown.
+
+Halyard configs can be used to inform Spinnaker of desired state, but whatever state is defined in Front50 will be used as the canonical source for plugins. 
+When a service starts up, it will ask Front50 what plugins need to be installed and which of those need to be enabled or disabled. Halyard’s role will only go as far as updating Front50’s repository of plugin metadata.
+
+Using Front50 as a source of truth will require extension of PF4J.
+
+```proto
+syntax = "proto3";
+
+// PluginManifest is used during registration of a plugin
+message PluginManifest {
+	// The name of the plugin, which is used as a fully 
+	// qualified identifier, including a namespace and 
+	// plugin name, ex: "netflix/fast-properties"
+	string name = 1;
+
+	// A short description of the plugin
+	// ex: "Provides rollout safety of runtime application 
+	// config via pipelines"
+	string description = 2;
+
+	// The name of the plugin author, ex: "Netflix"
+	string provider_name = 3;
+
+	// The semver plugin version, ex: "1.0.0"
+	string version = 4;
+
+	// A list of plugin dependencies. Initially this will be
+	// informational only. ex: "netflix/somedep>=1.x"
+	repeated string dependencies = 5;
+
+	// required_versions is a repeated string of service version	// requirements. e.g. "clouddriver >= 5.33.0"
+	repeated string required_versions = 6;
+
+	// A list of extension points that the plugin implements.
+	// e.g. "pipelineStage"
+	repeated string provided_capabilities = 7;
+
+	// Whether or not the plugin is enabled. An unset value 
+	// should be considered as true.
+	boolean enabled = 8;
+
+	// If the plugin is in-process and this flag is set to true,
+	// the plugin will be embedded into the process without
+	// isolation, allowing full modification of the application
+	// lifecycle. This is inherently unsafe and caution must be
+	// used setting this field.
+	boolean inprocess_unsafe = 9;
+}
+```
+
+#### Sending Plugin Configuration to Deck
+While Front50 will be the source of truth for which plugins are installed and enabled,  we can pass that data via `settings.js` today and transition to using Front50 when it is ready. 
+Using `settings.js` as an intermediate step allows us to collect feedback from selected users in order to improve the user experience in the future.
+
+We plan to prototype alternative options for Deck, such as altering webpack to support dynamic resolution, or introducing a standard container that can build Deck on startup with the configured plugins from Front50.
+
+### Protobuf as the Remote Contract
+As a departure from PF4J, plugin contracts will be defined as Protobuf messages (not services), so long as they are _remote-capable_. 
+If a plugin is in-process only, and there’s no expectation that it becomes available for remote plugins, it does not need to define its contract as protobuf.
+
+However, for any remotely-capable extension, its contract must be defined as protobufs, rather than exposed as simple POJOs or interfaces. 
+A key distinction is that we’re not specifying gRPC as the transport.  
+A remote transport may be Amazon SQS or Redis PubSub: not necessarily gRPC.
+
+### Technology Compatibility Kits
+_alias: TCKs_
+
+A convention of writing TCKs for each Extension Point will be established. 
+The concept of a TCK is borrowed from the JSR world, where compatibility tests are written which assert correct behavior of various features.  
+Each implementation must them pass those compatibility tests.  
+For Spinnaker plugins, TCKs can be written for two different consumers: Extension Point providers and Plugin developers.
+
+TCKs assert correct behavior of an extension point and its associated plugin(s). 
+For any extension point, a TCK will be written that asserts even pathological plugins behave in an expected way while integrating with the extension point. 
+For a plugin, a TCK will be provided that a plugin developer can use to assert correctness without necessarily needing to run other Spinnaker services.
+
+This goes further than the convention of using Protobuf to define the interfaces of a particular plugin, where Protobufs only define the interface, whereas a TCK will assert behavior that lies beneath the interface. 
+Combining a TCK and adherence to the interface, contracts will be fully asserted between a plugin and its extension point(s).
+
+### Up/Downgrading Plugins
+This section applies to federated development scenarios, where the operator is not necessarily the same as the plugin developers; where the deploy cadence of a plugin does not adhere to the deploy cadence of the core Spinnaker services.
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
+All in-process plugins should be downloadable. 
+When a service starts up, it will ask Front50 for a manifest of plugins, downloading all installed plugins that are not already present on the filesystem (in the case of baked-in plugins) to a locally-accessible directory. 
+At this point, PF4J will scan the directory, installing them into the process and enabling them based on the information pulled from Front50. 
+Should Front50 be unavailable, the service will fail to start.
+
+When a new plugin version is released and Front50 is updated with a new desired version, it is the responsibility of the operator to deploy this new version.
+_**TBD**: How will an operator know a new version exists?_ 
+This new version will need to be released in lock-step with the services that it is associated with. 
+
+Actual plugin artifacts can be stored at any HTTP-accessible endpoint. 
+Artifact repositories will support optional authentication, whereas checksums will always be required.
+There is no hard rule on what technology should be used for the artifact repository, but initial development will use basic HTTP endpoints.
+The download mechanism will be built into the PF4J integration within kork, either using [pf4j/pf4j-update](https://github.com/pf4j/pf4j-update) or a similar system that fits our needs.
+
+A tightly-coupled service and plugin deployment is a poor developer and operator story, but an acceptable shortcoming for an initial release.  There are no current plans for supporting a world where in-process plugins can be upgraded at runtime. There is a lot of complexity and a broad spectrum of potential bugs by hot-reloading plugins, so we aim to avoid this issue entirely by not supporting this use case. If true decoupling of service and plugin is desired, the plugin must be developed and deployed as a remote plugin.
+
+#### Remote
+Remote plugin upgrade process is not much different than iterating a remote service, since that’s what it is.
+
+When a remote plugin is released, its internal functionality will change immediately, just as it would be expected for a remote service. 
+When the remote plugin starts, it must send a registration event to Spinnaker. 
+This registration event would contain, at least:
+
+* Plugin ID
+* Other plugin metadata (author, description, version, etc).
+* What extension points it integrates with
+
+Upon receiving this registration event, Spinnaker would then update the Front50 registry with the new information and downstream services would re-bootstrap the plugin, adding or removing it from extension points as necessary.
+
+### Deployment of Remote Plugins
+Remote plugins can be deployed using whatever means the plugin developer desires.
+In the case of a remote plugin being developed by a third-party outside of the organization, the operator may need to deploy and maintain the plugin themselves.
+Halyard can be used for registering remote plugins with Front50.
+In the absence of Halyard, registering remote plugins with Front50 would be done through the API, just as Halyard would do.
+
+### Deployment of In-Process Plugins
+Downloading plugin resources is necessary for in-process plugins.
+Services need to have plugin resources available locally in order to load them into their process.
+Plugin resources are hosted on a remote server and must be downloaded in order to use them.
+The existing method of downloading plugin resources involves the service itself downloading plugins via a given URI.
+
+#### Kubernetes Environment
+Halyard can be used for deploying plugins to Kubernetes environments.
+To deploy plugins, Halyard will follow the same pattern as Kubernetes regarding downloading resources prior to runtime.
+Downloading of resources will be done through init containers and volume mounts that are then shared with their respective service.
+By using init containers to download resources, we can standardize on one method of deploying plugins to spinnaker services.
+Plugin resources will be placed in the following locations:
+
+- Java services: `/opt/spinnaker/plugins/resources`
+- Deck: `/opt/deck/html/plugins`
+
+Plugins are placed in a different location for Deck so that it can serve plugin resources.
+
+#### VM Based Environment
+Plugins and their configurations can be downloaded or baked into the image.
+
+### Plugin Configuration
+
+#### Remote Plugins
+Halyard or the operator, depending on preference, will be tasked with providing plugin configurations and plugin overrides to the remote plugins.
+
+#### In-Process Plugins
+PF4J plugins are configured with properties files and deployments are configured via `manifest.yml` file.
+Properties files configure the application and `manifest.yml` lets halyard know where plugin resources are located.
+`manifest.yml` can also override properties from properties file.
+PF4J will handle loading plugins into a Spinnaker service, while `manifest.yml` will inform which resources should be deployed to each service.
+
+### Dependencies
+A series of new kork modules will be provided for libraries and services alike to consume to expose new extension points. 
+Under the covers, this introduces PF4J as a new global service dependency.
+
+In-process plugins, by default, will have their own ClassLoader that is isolated from other Plugin ClassLoaders. 
+These Plugin ClassLoaders will be children of the root ClassLoader, meaning they cannot have dependencies that conflict with the root ClassLoader. 
+Future work may be performed to allow for different levels of isolation if necessary.
+
+**TBD**: We would like to offer different plugin runtimes besides in-process as plugins themselves. 
+These plugin runtimes would then come with their own dependencies that would be exposed to the parent service, such as gRPC or Lambda function invocation.
+
+## Drawbacks
+There are a handful major issues with the proposed direction:
+
+1. In-process plugins have a poor developer and operator experience which can negatively affect end-users, especially in rollback scenarios. 
+In-process plugins also require developers to have more knowledge of the Spinnaker toolchain and potentially the service architecture. 
+Exposing logging, telemetry and crash reports to plugin developers for in-process plugins will be difficult or impossible dependent on a Spinnaker installation’s preferences.
+2. Supporting alternative runtimes adds complexity to the final implementation and will likely require good judgment on an extension point developer to write the correct abstractions.
+
+## Prior Art and Alternatives
+The existing plugins code that has come out from [#4181](https://github.com/spinnaker/spinnaker/issues/4181)  is a successful implementation of plugins, but only supports an in-process runtime model, and is implementing a bespoke plugin system. 
+It would be preferential to use an existing library for plugins, where possible, that we don’t need to solely maintain.
+
+Furthermore, the existing system does not afford alternate plugin runtimes. 
+In-process plugins have the benefit of being capable of a wide variety of changes, but come at the cost of a poor operator story in that service and plugin deploy cadences must be in lock-step. 
+This model is fine for Spinnaker installations where the operator is also the plugin developer, but offers a poor story for environments where the plugin developer and Spinnaker service operators are not the same group and do not want to coordinate and be coupled to each other’s deploy cadences.
+
+The proposed plugin system will aim for feature parity of the existing plugin system, but will not guarantee backwards compatibility of developer or operator experience. 
+The priority is in satisfying broad-strokes business requirements, as opposed to satisfying existing technical capabilities.
+
+## Known Unknowns
+There are a lot of known unknowns. 
+It’s the intention that separate proposals will be made for many of these unknowns, such as development and publishing lifecycle, as well as operator experiences.
+
+* In-process plugins
+	* How will plugin authors get metrics, logs, or crash reports from their plugins?
+	* How will operators be notified that new plugin versions exist and can be released?
+	* Do we want to support updating backend plugins at runtime?
+* Deck plugins
+	* We want to experiment with some Webpack magic to achieve plugin loading, although we’re unsure if this will be possible yet.
+* Service interaction
+	* How do we handle behavior of faulty plugins?
+		* If a plugin brings a service down entirely, should there be some mechanism for tracking which plugin is at fault? How would this be exposed?
+
+## Security, Privacy, and Compliance
+In-process plugins present an interesting security issue. 
+Plugins can introduce new dependencies which expose CVEs, they can access internal APIs that may break core—or other plugins—behavior. 
+Further, a story around metrics, logs and error reporting for plugin developers becomes harder as there’s greater possibility for leaking potentially sensitive information to parties that do not necessarily have organizational approval for such data.
+
+Initially speaking, things like logs, metrics and crash reports for in-process plugins will go unaddressed, putting the sole burden of security judgment on the Operator.
+
+In-process plugins are not fully isolated, so close scrutiny of a plugin’s behavior is mandatory for any Operator prior to installation or enabling.
+
+## Operations
+This proposal alters how Spinnaker development will be performed, as well as how Spinnaker functionality is released and delivered to various environments. 
+We do not yet know the exact ways that this will change, however the goal is to make development of Spinnaker—for both core and extensions—easier.
+
+## Risks
+* There is an existing plugin system that has made assumptions on plugin loading. 
+There is risk that switching its internal functionality to PF4J may be in some ways backwards incompatible; such as how plugins are detected, or how config is loaded for them.
+	* This will likely cause breakages in existing plugins that will need to be migrated.
+* Service rollbacks when in-process plugins are being used will also rollback plugin versions, which may be undesired behavior.
+* Having Front50 as the source of truth for what plugins are installed across the system, and which ones are enabled or disabled introduces it as a single point of failure for a yet unknown, wide-sweeping segment of functionality for Spinnaker.
+	* Should Front50 be unavailable, services should use the shipped / baked config as defaults.
+	* Gate should store an in-memory cache of the desired plugin state such that Deck may have less dependence on a healthy Front50 deployment.


rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+	// A short description of the plugin
+	// ex: "Provides rollout safety of runtime application 
+	// config via pipelines"
+	string description = 3;
+
+	// The name of the plugin author, ex: "Netflix"
+	string provider_name = 4;
+
+	// The semver plugin version, ex: "1.0.0"
+	string versionn = 5;
```
marchello2000:
```suggestion
	string version = 5;
```

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+	// A list of plugin dependencies. Initially this will be
+	// informational only. ex: "netflix/somedep>=1.x"
+	repeated string dependencies = 6;
+
+	// required_versions is a repeated string of service version	// requirements. e.g. "clouddriver >= 5.33.0"
+	repeated string required_versions = 7;
+
+	// A list of extension points that the plugin implements.
+	// e.g. "pipelineStage"
+	repeated string provided_capabilities = 8;
```
marchello2000: I think the capabilities be `enum`s

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+This goes further than the convention of using Protobuf to define the interfaces of a particular plugin, where Protobufs only define the interface, whereas a TCK will assert behavior that lies beneath the interface. Combining a TCK and adherence to the interface, contracts will be fully asserted between a plugin and its extension point(s).
+
+### Up/Downgrading Plugins
+This section applies to federated development scenarios, where the operator is not necessarily the same as the plugin developers; where the release cadence of a plugin does not adhere to the release cadence of the core Spinnaker services.
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
+When a service starts up, PF4J will scan a designated locally-accessible directory for plugins to install (these plugins may be downloaded at runtime or baked into the image at built-time). Once an application has started, it will notify Front50 of the plugins that are installed, and get a response dictating which plugins are to be enabled.
```
marchello2000: can you elaborate on why this order of operations?
e.g. I would expect, given that Front50 is source of truth, that we call Front50 first, then try to load the enabled plugins. This way the flow is the same of downloaded/not downloaded plugins

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+In the case of a runtime-downloaded plugin, version to download will be dictated by Front50. Should Front50 be unavailable, the service will download the version defined by the service, or if none is defined, the latest plugin version that matches the Spinnaker service version.
```
marchello2000: is this complexity needed? Why not, at least in the beginning, say no front50 - no plugins.

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+Tightly-coupled service and plugin releases is a poor developer and operator story, but an acceptable shortcoming for an initial release. By using Front50 as the source of truth, we’ll be fairly well equipped to make an acceptable runtime plugin loading story.
+
+Assuming a world where in-process plugins can be upgraded at runtime, services will need a background process to download these new plugins whenever a delta is detected in Front50, and the plugin refreshed. PF4J offers an abstraction for this reloading mechanism, but we would need to figure out how Spring integrations are handled. For example, it’s possible we may need to make the constraint that exported beans from a plugin are not refreshed on plugin up/down grade at least initially.
+
+#### Remote
+Remote plugin upgrade process is not much different than iterating a remote service, since that’s what it is.
+
+When a remote plugin is released, its internal functionality will change immediately, just as it would be expected for a remote service. When the plugin starts, it must send a registration event to Spinnaker. This registration event would contain, at least:
+
+* Plugin ID
```
marchello2000: why not send the same manifest as https://github.com/spinnaker/community/pull/43/files#diff-d01483cbab3802f88083e6b1541851d2R115

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+])
+class PipelineStageExtension : ExtensionPoint {
+	// ...
+}
+```
+
+It is highly recommended to read the documentation of PF4J to fully understand the scope of this proposal.
+
+### Front50 as the Source of Truth
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. Whether plugins are implemented as in-process or not, they Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled.
```
srekapalli: they/the Front50.. ( small typo )

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+It is highly recommended to read the documentation of PF4J to fully understand the scope of this proposal.
+
+### Front50 as the Source of Truth
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. Whether plugins are implemented as in-process or not, they Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled. 
+
+The Front50 plugin registry will also be exposed through Gate, and can be consumed by Deck to load the correct plugins on application startup.
+
+Plugins can also be enabled or disabled at runtime without redeploying a service. This flag is sourced from Front50, which all services will be listening for change events from. Whether this change event is published or polled for initially is unknown.
+
```
srekapalli: Plugin management can be a component with in Deck that will allow operators to turn on / off .. may be ?

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+TCKs satisfy asserting correct behavior of an extension point and its associated plugin(s). For any extension point, a TCK will be written that asserts even pathological plugins behave in an expected way while integrating with the extension point. For a plugin, a TCK will be provided that a plugin developer can use to assert correctness without necessarily needing to run other Spinnaker services.
+
+This goes further than the convention of using Protobuf to define the interfaces of a particular plugin, where Protobufs only define the interface, whereas a TCK will assert behavior that lies beneath the interface. Combining a TCK and adherence to the interface, contracts will be fully asserted between a plugin and its extension point(s).
+
+### Up/Downgrading Plugins
+This section applies to federated development scenarios, where the operator is not necessarily the same as the plugin developers; where the release cadence of a plugin does not adhere to the release cadence of the core Spinnaker services.
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
```
marchello2000: I think we should address the behavior/handling of faulty plugins.
E.g. at a minimum, we should use 
1. our top-level exception handlers, and inspect unhandled exceptions for those coming from the plugins
2. If a plugin happens to completely bring down a Spinnaker service, say 3 times in a row, we should disable said plugin. If it's not possible to tell which plugin was the cause, disable all plugins.

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+The proposed plugin system will aim for feature parity of the existing plugin system, but will not guarantee backwards compatibility of developer or operator experience. The priority is in satisfying broad-strokes business requirements, as opposed to satisfying existing technical capabilities.
+
+## Known Unknowns
+There are a lot of known unknowns. It’s the intention that separate proposals will be made for many of these unknowns, such as development and publishing lifecycle, as well as operator experiences.
+
+* In-process plugins
+	* How will plugin authors get metrics, logs, or crash reports from their plugins?
+	* How will operators be notified that new plugin versions exist and can be released?
+	* Do we want to support updating backend plugins at runtime?
```
marchello2000: not sure how to capture this... but say a plugin extends stages, the plugin was upgraded in a non-backwards compatible way but the messages in the queue/stage contexts are still in flight so the new plugin can't pickup where the old one left off. Obv, a burden on the plugin developer but we should make it clear in docs/maybe think of tck testing for this?

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+message PluginManifest {
+	// The name of the plugin, which is used as a fully 
+	// qualified identifier, including a namespace and 
+	// plugin name, ex: "netflix/fast-properties"
+	string name = 1;
+
+	// A short description of the plugin
+	// ex: "Provides rollout safety of runtime application 
+	// config via pipelines"
+	string description = 3;
```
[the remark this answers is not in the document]
robzienert: I deleted it and then was like, "Man, I'm not updating that, it's just some markdown." 😬

rfc/plugins.md
robzienert: I would rather not. Making it an enum would couple the proto to the services more than I'd prefer, and would prevent `-nflx` style extensions from providing plugins of their own without modifying the OSS proto.

rfc/plugins.md
robzienert: Because there's the possibility that Front50 may be down. This isn't often, but if plugins existed that provided critical functionality, and then while a service was starting, Front50 was unavailable for a moment, we'd lose the critical functionality entirely.

rfc/plugins.md
robzienert: It probably would. This section hasn't been updated since I added that proto.

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+In-process plugins, by default, will have their own ClassLoader that is isolated from other Plugin ClassLoaders. These Plugin ClassLoaders will be children of the root ClassLoader, meaning they cannot have dependencies that conflict with the root ClassLoader. Future work may be performed to allow for different levels of isolation if necessary.
+
+**TBD**: We would like to offer different plugin runtimes besides in-process as plugins themselves. These plugin runtimes would then come with their own dependencies that would be exposed to the parent service, such as gRPC or Lambda function invocation.
+
+## Drawbacks
+There are a handful major issues with the proposed direction:
+
+1. In-process plugins have a poor developer and operator experience which can negatively affect end-users, especially in rollback scenarios. In-process plugins also require developers to have more knowledge of the Spinnaker toolchain and potentially the service architecture. Exposing logging, telemetry and crash reports to plugin developers for in-process plugins will be difficult or impossible dependent on a Spinnaker installation’s preferences.
```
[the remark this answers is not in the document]
robzienert: Yes, but that would first require a bugsnag plugin. Where does bugsnag send its data if the plugin isn't working correctly?

rfc/plugins.md
robzienert: Yes. I think we'll definitely want an operator view in Deck that would allow operators to see what plugins are available, installed, enabled/disabled and be able to actuate the ones that are installed (enable/disable as needed). My intention (not mentioned in this proposal) is that plugin management would be implemented as a "core" plugin to dogfood the whole process.

robzienert: Ideally (medium-/long-term), I'd like to make Fast Properties a plugin and then build the operator experience atop that.

rfc/plugins.md
robzienert: I was thinking that plugins would need to be written in a backwards-compatible way. Plugin manifests will be able to define a minimum spinnaker version, but I had not addressed backwards-incompatible plugin behavior. This seems like something that falls on the plugin developer to manage, rather than something that we dictate and enforce via the core platform? WDYT?

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+This goes further than the convention of using Protobuf to define the interfaces of a particular plugin, where Protobufs only define the interface, whereas a TCK will assert behavior that lies beneath the interface. Combining a TCK and adherence to the interface, contracts will be fully asserted between a plugin and its extension point(s).
+
+### Up/Downgrading Plugins
+This section applies to federated development scenarios, where the operator is not necessarily the same as the plugin developers; where the release cadence of a plugin does not adhere to the release cadence of the core Spinnaker services.
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
+When a service starts up, PF4J will scan a designated locally-accessible directory for plugins to install (these plugins may be downloaded at runtime or baked into the image at built-time). Once an application has started, it will notify Front50 of the plugins that are installed, and get a response dictating which plugins are to be enabled.
```
christopherthielen: For the "download at runtime" (just in time) scenario, do we leave this up to operators to figure out, or do we think there will be a standard mechanism?

rfc/plugins.md
marchello2000: hm... that's fair, thanks for the explanation.
I guess we could have a "validation" mechanism of sorts for developers before they publish, if we wanted to

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+When a new plugin version is released, it is the responsibility of the operator to release this new version. _**TBD**: How will an operator know a new version exists?_ This new version will need to be released in lock-step with the services that it is associated with. In the future, this functionality _may_ be dynamically sourced and then re-activated. See [pf4j/pf4j-update](https://github.com/pf4j/pf4j-update)
+
+Tightly-coupled service and plugin releases is a poor developer and operator story, but an acceptable shortcoming for an initial release. By using Front50 as the source of truth, we’ll be fairly well equipped to make an acceptable runtime plugin loading story.
+
+Assuming a world where in-process plugins can be upgraded at runtime, services will need a background process to download these new plugins whenever a delta is detected in Front50, and the plugin refreshed. PF4J offers an abstraction for this reloading mechanism, but we would need to figure out how Spring integrations are handled. For example, it’s possible we may need to make the constraint that exported beans from a plugin are not refreshed on plugin up/down grade at least initially.
```
christopherthielen: This part give me the heebie jeebies

rfc/plugins.md
marchello2000: hm... if it's to account for momentary blips in front50 wouldn't it be better to retry/fail service start if we can't load the critical plugins rather than potentially loading a potentially wrong version of the plugin?

rfc/plugins.md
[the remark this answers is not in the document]
marchello2000: if which plugin isn't working? the bugsnag client not working? That's like saying what if elk/sumo is down, no? BugSnag (and similar services) have a ton resilience and store data locally until it can be sent, etc. but yes, the spinnaker plugin developer would have to load bugsnag/or equivalent and if the plugin fails early on it might not have initialized bugsnag yet...

rfc/plugins.md
marchello2000: Yeah, I agree. I am just thinking how to we document/inform the developers that this is required/how do we present to the operators/end-users when a plugin fails vs. spinnaker fails. I think we should make sure that failure attribution is crystal clear

rfc/plugins.md
robzienert: Yeah, I can get behind this flow. Thank you both.

rfc/plugins.md
robzienert: I'm going to follow the flow directed by https://github.com/spinnaker/community/pull/43#discussion_r333649332

Would this address your feedback?

rfc/plugins.md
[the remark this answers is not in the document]
robzienert: After thinking on this more, I like the essence of what you're getting at. Perhaps there's room for a diagnostics plugin of sorts, if you want to call it that, that can be implemented to report telemetry, logs and errors to a separate system. The tricky bit here is each plugin developer is going to want to consume this information in a way that suits them; which would mean varying diagnostics plugins floating around depending on the various plugins that are enabled.

Am I going off the rails here? Is there merit in taking this thought further?

rfc/plugins.md
robzienert: That's a good thing to keep in mind. Sounds like a plugin developer documentation thing. Thank you!

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+It is highly recommended to read the documentation of PF4J to fully understand the scope of this proposal.
+
+### Front50 as the Source of Truth
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. Whether plugins are implemented as in-process or not, they Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled. 
+
+The Front50 plugin registry will also be exposed through Gate, and can be consumed by Deck to load the correct plugins on application startup.
+
+Plugins can also be enabled or disabled at runtime without redeploying a service. This flag is sourced from Front50, which all services will be listening for change events from. Whether this change event is published or polled for initially is unknown.
+
+Halyard configs can be used to inform services of their _initial state_. When a service starts up, it will notify Front50 of what in-process plugins are installed, however Front50 will be responsible for determining the enabled state.
```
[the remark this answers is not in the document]
robzienert: I will update. Here's the difference:

A plugin is "installed" if it's in the system. An installed plugin can either be in an "enabled" or "disabled" state. In a disabled state, it will not be getting invoked and will not have its internal components started. If the plugin is transitioning state from enabled to disabled (or vice-versa), the plugin framework calls a correlated method to handle whatever logic the plugin must step through to become running or stopped.

I will update the doc to cover this lifecycle. It's dictated (mostly) by PF4J but will have some deviations based on our decision to use Front50 as SOT.

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Up/Downgrading Plugins
+This section applies to federated development scenarios, where the operator is not necessarily the same as the plugin developers; where the release cadence of a plugin does not adhere to the release cadence of the core Spinnaker services.
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
+When a service starts up, PF4J will scan a designated locally-accessible directory for plugins to install (these plugins may be downloaded at runtime or baked into the image at built-time). Once an application has started, it will notify Front50 of the plugins that are installed, and get a response dictating which plugins are to be enabled.
+
+In the case of a runtime-downloaded plugin, version to download will be dictated by Front50. Should Front50 be unavailable, the service will download the version defined by the service, or if none is defined, the latest plugin version that matches the Spinnaker service version.
```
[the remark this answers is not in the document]
robzienert: If it's an in-process plugin, I think we can always download, just as I think any in-process plugin can be burnt into an image. I'll alter the language to make this clear.

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+	// A short description of the plugin
+	// ex: "Provides rollout safety of runtime application 
+	// config via pipelines"
+	string description = 3;
+
+	// The name of the plugin author, ex: "Netflix"
+	string provider_name = 4;
+
+	// The semver plugin version, ex: "1.0.0"
+	string version = 5;
```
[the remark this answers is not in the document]
robzienert: This is resolved by the `required_versions` field. If an extension point changes, that will be reflected via the service version.

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+It is highly recommended to read the documentation of PF4J to fully understand the scope of this proposal.
+
+### Front50 as the Source of Truth
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. Whether plugins are implemented as in-process or not, Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled.
+
+The Front50 plugin registry will also be exposed through Gate, and can be consumed by Deck to load the correct plugins on application startup.
+
+Plugins can also be enabled or disabled at runtime without redeploying a service. This flag is sourced from Front50, which all services will be listening for change events from. Whether this change event is published or polled for initially is unknown.
+
+Halyard configs can be used to inform services of their _initial state_. When a service starts up, it will notify Front50 of what in-process plugins are installed, however Front50 will be responsible for determining the enabled state.
```
link108: I don't think halyard configs should be used to inform initial state, but rather desired state. With Front50 being the source of truth, I think halyard configs are relevant for determining which plugins are present/enabled in Front50. 

I'd expect halyard would be the tool to add plugin info into Front50, but I am open to other ideas as well.  I view halyard as the standard way of deploying spinnaker, but I do understand not everyone uses it to deploy. Ideally I'd like to prevent a sprawl of Spinnaker tooling, as that would increase barrier of entry, usability, etc, but I do want to be considerate of users who do not use or want to use halyard.

rfc/plugins.md
link108: Ah this is new news to me! I was under the impression we were trying to move away from services downloading plugin resources, is that not the case anymore? 

From my experience, most/all people running spinnaker are also running a secure artifact store, eg artifactory. I would expect plugin resources would be stored there and referenced from Front50, so having Front50 be the source of truth for where plugins are loaded is good with me.

rfc/plugins.md
link108: as a spinnaker operator, if I expect a plugin to be loaded and its not, I'd like the deploy to fail or retry, as the desired state has not been achieved. This is particularly important and would be confusing if I deploy a new version of a plugin and the new functionality is not present.

rfc/plugins.md
robzienert: Having a download story seems a necessity if we want to decouple the release cycles of the Spinnaker services and their plugins. We're not yet going to address runtime updates or anything of the sort, so it's not going to be as flexible as a remote plugin, but it's something.

Front50 won't be the artifact store, but just the storage that knows what plugins are installed & enabled. Artifactory, some blob store, or whatever will still be needed to actually download the artifacts from.

rfc/plugins.md
robzienert: Yes, agreed.

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Multiple Plugin Runtimes
+There’s a place and a time for different plugin runtimes. In-process extensibility offers a lot of power and network efficiency at the cost of higher coupling to core service rollouts, whereas a remote plugin runtime trades runtime efficiency and reliability for development and release efficiency.
+
+The goal of this proposal is to offer the best of both worlds by allowing extension point developers (Core Spinnaker developers) to decide what runtime(s) are the best for the extension.
+
+Extension points will be defined and exposed via PF4J.
+
+### PF4J as a Foundation
+[PF4J](https://github.com/pf4j/pf4j) is to be used as the foundation for all plugin functionality. While designed for in-process extensibility, we’ll be using PF4J to define both the contract for in-process plugins, as well as remote plugins. Remote plugins would be supported in two parts: First, by including a plugin that activates a new plugin runtime (e.g. gRPC, or a Docker co-process). Second, a remote plugin stub that uses the runtime plugin as a transport.
```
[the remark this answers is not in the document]
luispollo: I had a similar question. If I understand correctly, I think it's worth clarifying that, while remote plugins would still require an in-process counterpart, it's a one-time effort to onboard the protocol-enablement in-process plugin (what you called the "plugin runtime" above) which can then support multiple different remote plugins using that same protocol. I was a bit confused with the "remote plugin stub" description.

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Design
+### Taxonomies
+| Term | Definition |
+| - | - |
+| **Extension Point** | A strongly typed service- or library-defined point of extension |
+| **Plugin** | An implementation of an Extension Point |
+
+### Multiple Plugin Runtimes
+There’s a place and a time for different plugin runtimes. In-process extensibility offers a lot of power and network efficiency at the cost of higher coupling to core service rollouts, whereas a remote plugin runtime trades runtime efficiency and reliability for development and release efficiency.
```
luispollo: >  whereas a remote plugin runtime trades runtime efficiency and reliability for development and release efficiency.

A remote model also allows developer to pick their programming language/framework/runtime of choice.

rfc/plugins.md
luispollo: Remote monitors for the Monitored Deploy Strategy seem like a good candidate as well. It's unlucky timing that we already have an HTTP-based API contract defined at this point that is very specific to that use case, but thought I'd suggest it anyway.

luispollo: One more thought on plugin management: it's impossible not to think of Jenkins while reading through this, and both the flexibility that its plugin model allows and the nightmare that is managing plugins in a Jenkins installation... I wondered if we should call out whether managing plugin updates specifically are a goal or non-goal.

luispollo: > Remote monitors for the Monitored Deploy Strategy seem like a good candidate as well. It's unlucky timing that we already have an HTTP-based API contract defined at this point that is very specific to that use case, but thought I'd suggest it anyway.

P.S. Or maybe deployment strategies in general? Interested in your thoughts on where to draw the line here since you proposed stages as a candidate for plugins but I'm unclear whether it would make sense to replace an entire Deploy stage if you want to use a different deployment strategy?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+	string provider_name = 4;
+
+	// The semver plugin version, ex: "1.0.0"
+	string version = 5;
+
+	// A list of plugin dependencies. Initially this will be
+	// informational only. ex: "netflix/somedep>=1.x"
+	repeated string dependencies = 6;
+
+	// required_versions is a repeated string of service version	// requirements. e.g. "clouddriver >= 5.33.0"
```
luispollo: "repeated string" as in a comma-separated list?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+	// config via pipelines"
+	string description = 3;
+
+	// The name of the plugin author, ex: "Netflix"
+	string provider_name = 4;
+
+	// The semver plugin version, ex: "1.0.0"
+	string version = 5;
+
+	// A list of plugin dependencies. Initially this will be
```
luispollo: Format? Comma-separated list?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+	// informational only. ex: "netflix/somedep>=1.x"
+	repeated string dependencies = 6;
+
+	// required_versions is a repeated string of service version	// requirements. e.g. "clouddriver >= 5.33.0"
+	repeated string required_versions = 7;
+
+	// A list of extension points that the plugin implements.
```
luispollo: Do we need to provide any additional info on how names declared here map to the extension point definitions which are declared in code?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+	// Whether or not the plugin is enabled. An unset value 
+	// should be considered as true.
+	boolean enabled = 9;
+}
+```
+
+#### Sending Plugin Configuration to Deck
+
+While Front50 will be the source of truth for which plugins are enabled, we can pass that data via settings.js today and transition to using Front50 when it is ready.
```
luispollo: > we can pass that data via settings.js today
Maybe rewrite as "we can pass that data via Deck's configuration file (`settings.js`) today" in case readers are unfamiliar with Deck's internals?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+Using settings.js as an intermediate step allows us to collect feedback from selected users in order to improve the user experience in the future.
+
+Replacing/Updating the templating format for settings.js is out of scope for this project, so we propose to pass that data via a string that Deck can parse to determine which Plugins are installed.
+
+### Protobuf as the Remote Contract
+As a departure from PF4J, plugin contracts will be defined as Protobuf messages (not services), so long as they are _remote capable_. If a plugin is in-process only, and there’s no expectation that it becomes available for remote plugins, it does not need to define its contract as protobuf.
```
luispollo: This was a bit confusing to me, given the emphasis on P4J earlier in the doc, as well as the plugin manifest definition above in protobuf format (which, if I understand correctly, has nothing to do with the remote contract definition you speak of here). In any case, I think if we actually had the protocol (request/response?) message definitions here that would make things clearer.

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+However, for any remotely-capable extension, its contract must be defined as protobufs, rather than exposed as simple POJOs or interfaces. A key distinction is that we’re not specifying gRPC as the transport.  A remote transport may be Amazon SQS or Redis PubSub, not necessarily gRPC.
+
+### Technology Compatibility Kits
+_alias: TCKs_
+
+A convention of writing TCKs for each Extension Point will be established. The concept of a TCK is borrowed from the JSR world, where compatibility tests are written which assert correct behavior of various features.   Each implementation must them pass those compatibility tests.  For Spinnaker plugins, TCKs can be written for two different consumers: Extension Point providers and Plugin developers.
```
luispollo: This is cool! 👏 👏 👏

rfc/plugins.md
luispollo: > One more thought on plugin management: it's impossible not to think of Jenkins while reading through this, and both the flexibility that its plugin model allows and the nightmare that is managing plugins in a Jenkins installation... I wondered if we should call out whether managing plugin updates specifically are a goal or non-goal.

I see the section on upgrades/downgrades down below now. I'd suggest adding that to the goals list.

rfc/plugins.md
luispollo: Do we need to include anything on expectations for supported download protocols and authentication mechanisms?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
+When a service starts up, PF4J will scan a designated locally-accessible directory for plugins to install (these plugins may be downloaded at runtime or baked into the image at built-time). Once an application has started, it will notify Front50 of the plugins that are installed, and get a response dictating which plugins are to be enabled.
+
+In the case of a runtime-downloaded plugin, version to download will be dictated by Front50. Should Front50 be unavailable, the service will download the version defined by the service, or if none is defined, the latest plugin version that matches the Spinnaker service version.
+
+When a new plugin version is released, it is the responsibility of the operator to release this new version. _**TBD**: How will an operator know a new version exists?_ This new version will need to be released in lock-step with the services that it is associated with. In the future, this functionality _may_ be dynamically sourced and then re-activated. See [pf4j/pf4j-update](https://github.com/pf4j/pf4j-update)
```
luispollo: > When a new plugin version is released, it is the responsibility of the operator to release this new version.

"...the responsibility of the operator to _install_ this new version"?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+Tightly-coupled service and plugin releases is a poor developer and operator story, but an acceptable shortcoming for an initial release. By using Front50 as the source of truth, we’ll be fairly well equipped to make an acceptable runtime plugin loading story.
```
luispollo: I'm confused by this statement. Why do we need to tightly couple service and plugin releases if we're proposing to download plugins at runtime? If a new version of the plugin is released and does not require an upgraded core component, couldn't we just download and install the new version of the plugin?

rfc/plugins.md
luispollo: > _**TBD**: How will an operator know a new version exists?

The Jenkins model is to have an official plugin repository from which metadata about available plugins can be queried and displayed on the UI. Not that we want to follow Jenkins 😃, but it occurred to me that mentioning a plugin repository as a goal/non-goal might be relevant as well.

rfc/plugins.md
luispollo: Since, according to this, we only get updated plugin info from the remote endpoint only upon plugin start, there's a high likelihood that the remote may be updated and we'll not refresh info in Front50 until the next time the service starts. Is that important enough that we should we poll periodically?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+* Plugin ID
+* Other plugin metadata (author, description, version, etc).
+* What extension points it integrates with
+
+Upon receiving this registration event, Spinnaker would then update the Front50 registry with the new information and downstream services would re-bootstrap the plugin, adding or removing it from extension points as necessary.
+
+### Deployment of Remote Plugins
+
+Halyard based Spinnaker installations will use halyard to deploy remote plugins, similar to other spinnaker services.
```
luispollo: Oooh, that sounds pretty ambitious. One of the benefits of remote plugins is that they can be coded in whatever stack the developer prefers. I'm unfamiliar with the magic powers of halyard, but I'd guess it's tailored to deploying JVM-based services. How do we plan to account for other types of runtimes?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+The existing method of downloading plugin resources involves the service itself downloading plugins via a given URI.
+While this works for the java services, it is not a general solution.
+Furthermore, downloading of plugin resources is a deployment concern and should not be part of the runtime environment.
+
+#### Kubernetes Environment
+
+Halyard can be used for deploying plugins to Kubernetes environments.
+To deploy plugins, Halyard will follow the same pattern as kubernetes regarding downloading resources prior to runtime.
+Downloading of resources will be done through init containers and volume mounts that are then shared with their respective service.
+By using init containers to download resources, we can standardize on one method of deploying plugins to spinnaker services.
```
luispollo: Hmm... I can see the value in this, but if we're going to support runtime downloads as discussed above, there will need to be some logic in the core service(s) to do that, so we'll end up with two different ways to download and install plugins, no? Do we plan to start without runtime download support and add it later?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+
+Plugins and their configurations are baked into image.
+
+### Plugin Configuration
+
+#### Remote Plugins
+
+Halyard or the operator, depending on the preference, will be tasked with providing plugin configurations and plugin overrides to the remote plugins.
+
+#### In Process Plugins
```
luispollo: For consistency:
```suggestion
#### In-Process Plugins
```

rfc/plugins.md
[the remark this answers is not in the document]
luispollo: FWIW, I think the plugin developer vs. the Spinnaker operator needs are very different here. What you guys discuss above makes sense for developers. For operators, I would expect all plugins to tap into the existing logging/monitoring infrastructure used by Spinnaker core components, in other words, it shouldn't any burden just because you added plugins to the mix.

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+Initially speaking, things like logs, metrics and crash reports for in-process plugins will go unaddressed, putting the sole burden of security judgment on the Operator.
+
+In-process plugins are not fully isolated, so close scrutiny of a plugin’s behavior is mandatory for any Operator prior to installation or enabling.
+
+## Operations
+This proposal alters how Spinnaker development will be performed, as well as how Spinnaker functionality is released and delivered to various environments. We do not yet know the exact ways that this will change, however the goal is to make development of Spinnaker—for both core and extensions—easier.
+
+## Risks
+* There is an existing plugin system that has made assumptions on plugin loading. There is risk that switching its internal functionality to PF4J may be in some ways backwards incompatible; such as how plugins are detected, or how config is loaded for them.
+	* This will likely cause breakages in existing plugins that will need to be migrated.
```
luispollo: Do we have an idea how many plugins have been developed against the old system?

rfc/plugins.md
```diff
@@ -0,0 +1,300 @@
...
+* Service rollbacks when in-process plugins are being used will also rollback plugin versions, which may be undesired behavior.
```
luispollo: Not if we're using Front50 as the source of truth, right? I thought service-defined plugin versions were only a fallback from retrieving the data from Front50 per above?

rfc/plugins.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Goals and Non-Goals
+* Goals
+	* Support plugins on both the frontend & backend services
+	* Support plugins spanning multiple services
+	* Support in-process and RPC-backed plugins for backend services
+	* Provide tooling for developers and operators
+* Non-goals
+	* Support for alternative runtimes (e.g. Lambda, Docker; to be addressed later)
+	* Dictation of what can or cannot be a plugin within Spinnaker
```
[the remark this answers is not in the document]
robzienert: Changed to `	* Identification of all extension points within Spinnaker`

rfc/plugins.md
robzienert: How about this?

> Halyard configs can be used to inform Spinnaker of desired state, but whatever state is defined in Front50 will be used as the canonical source for plugins. When a service starts up, it will ask Front50 what plugins need to be installed and which of those needs to enabled or disabled. Halyard’s role will only go as far as updating Front50’s repository of plugin metadata.

rfc/plugins.md
robzienert: It says `repeated string` ...

rfc/plugins.md
robzienert: https://developers.google.com/protocol-buffers/docs/proto#specifying-field-rules

rfc/plugins.md
robzienert: 🤷‍♂ We don't know yet.

rfc/plugins.md
robzienert: When the remote plugin starts, it will send its metadata to Spinnaker. This is a push model, where the remote plugin is responsible for telling Spinnaker what its metadata is. In this scenario, there is no likelihood of staleness assuming the plugin's push does not fail. The plugin can further decide to push this metadata to Spinnaker whenever it wants as well.

rfc/plugins.md
robzienert: I've removed this statement - there's no technical benefit to hammering remote plugins into a bespoke Spinnaker-only config tool. Not sure how this made it into the proposal. 😬

rfc/plugins.md
robzienert: Yes, we'll have two different ways to get plugins in a K8s environment. Unfortunately, this is pretty well out of our control at this point (@link108 keep me honest?). Runtime download support will not be initially supported (because we need to build it), but it will be an early feature as it's critical to how Deck's plugin model will work.

rfc/plugins.md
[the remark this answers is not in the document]
robzienert: Yes, absolutely, in-process plugins will be plugged into the existing telemetry / logging system, but that's not useful for the plugin developers: Take for example Netflix: No one outside of Delivery Engineering has access to the Spinnaker logs. We don't want to be in the business of copy/pasting logs to plugin developers whenever something goes sideways for something they developed.

rfc/plugins.md
robzienert: A very small amount. It's not something to lose sleep over, but this number will increase the longer we wait to provide a longer-term solution.

rfc/plugins.md
robzienert: If the past server group was disabled and we rollback to it, enabling a server group does not bounce the process or anything like that. If the rollback were to re-create the previous server group, then the latest plugins would be installed.

rfc/plugins.md
robzienert: I decided to call out the fact that Front50 being absent will fail the service startup process, but anything else on faulty plugins will be left as an exercise for discovery in Known Unknowns.

rfc/plugins.md
luispollo: Sorry. Unfamiliar with specifics of protobuf...

rfc/plugins.md
luispollo: Yes, yes... RTFM! 😄

rfc/plugins.md
luispollo: Ah, this was not clear at all to me when I was reading that... I assumed you were talking about the in-process counterpart to the remote plugin here, not the actual remote service. I guess maybe if we added something to indicate that "Spinnaker will expose an API for the remote plugin to send this registration event" or maybe that "the TCK will provide registration functionality to plugin developers" (if we intend to do that, but I didn't think we did for remote plugins).

rfc/plugins.md
link108: LGTM 👍

rfc/plugins.md
link108: I wouldn't say its set in stone, I'd prefer to have one method of deploying plugins. I'm all for going the remote downloadable route, with plugin metadata stored in Front50 👍 After some of the PR updates, I can more easily grok the remote-downloadable solution and can update this accordingly.

I do think Halyard still has a part to play in this, I view plugins as part of a spinnaker runtime and as such, should be part of the hal configs. Rather than deploy the plugins, Halyard would update Front50 with the plugin metadata, provided in the manifest.yml file.

rfc/plugins.md
link108: Ah maybe the statement was too broad, I should have specified this is for deploying container based remote plugins in a Kubernetes environment. The statement here was based on that, as Halyard deploys Spinnaker services as containers in Kubernetes environments. 

I mention it in another comment, but I view plugins as part of a Spinnaker runtime environment and Halyard should ensure the runtime environment is as configured. I do think Halyard should deploy container based plugins and should validate non-container based remote plugins are reachable/ready for traffic (eg similar to cloud provider account validation). 

I expect most remote plugin providers will distribute their plugins as containers, as is so very popular in the devops space these days, but time will tell. Do you guys have a similar expectation? Curious about your guys' thoughts on all this.

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+
+	// The name of the plugin author, ex: "Netflix"
+	string provider_name = 3;
+
+	// The semver plugin version, ex: "1.0.0"
+	string version = 4;
+
+	// A list of plugin dependencies. Initially this will be
+	// informational only. ex: "netflix/somedep>=1.x"
+	repeated string dependencies = 5;
```
link108: going off the example, I'm assuming that would be referencing a registry of some sort? If not, how are plugin resources referenced (eg the location of the jars in an artifact store)? 

Also, do we want to make any distinction between dependencies for different services (eg Orca needs x.jar, while Deck needs y.js, z.png)? or is each service responsible for downloading all of the plugin resources? I think we could also register a plugin multiple times if it spans multiple services, but that leads to complexity

link108: riffing off this some more, we may want to add a type as well. This is more related to Deck, but I can see Deck needing to load js and css files in different ways, but that hasn't really been though out too much. 

I'm open to discussing this in another design doc, I'm not sure how much we want to grow the scope of this design, but a plugin frontend component is pretty important piece.

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. 
+Whether plugins are implemented as in-process or not, Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled.
+
+The Front50 plugin registry will also be exposed through Gate, and can be consumed by Deck to load the correct plugins on application startup.
+
+Plugins can also be enabled or disabled at runtime without redeploying a service. 
+This flag is sourced from Front50, which all services will be listening for change events from. Whether this change event is published or polled for initially is unknown.
+
+Halyard configs can be used to inform Spinnaker of desired state, but whatever state is defined in Front50 will be used as the canonical source for plugins. 
+When a service starts up, it will ask Front50 what plugins need to be installed and which of those needs to enabled or disabled. Halyard’s role will only go as far as updating Front50’s repository of plugin metadata.
```
marchello2000: grammar: i assume: 
need~s~ to *be* enabled or disabled

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+	// the plugin will be embedded into the process without
+	// isolation, allowing full modification of the application
+	// lifecycle. This is inherently unsafe and caution must be
+	// used setting this field.
+	boolean inprocess_unsafe = 9;
+}
+```
+
+#### Sending Plugin Configuration to Deck
+While Front50 will be the source of truth for which plugins are installed and enabled, we can pass that data via Deck’s configuration file (`settings.js`), we can pass that data via `settings.js` today and transition to using Front50 when it is ready.
```
marchello2000: grammar: there is some repeatedness/confusionness happening here, i think you can scratch the first ~, we can pass that data via Deck’s configuration file (`settings.js`)~

rfc/plugins.md
robzienert: Taxonomy thing: I've been trying to say that the plugin registry is Front50, in that plugins are registered into it. Plugin repositories would be where plugins are stored for download. Does this seem a reasonable distinction?

So, I think the `dependencies` property doesn't need to specify the repository that a plugin is found in. Kork should be capable of configuring a list of 1+n repositories that it will look in (ordered) to find a plugin (in this example, `netflix/somedep>=1.x`). Check out [pf4j-update](https://github.com/pf4j/pf4j-update) for inspiration here. Similar to the `pf4j-spring` integration, I don't see this `pf4j-update` library totally hitting the right mark for us, but you get the idea.

As for distinction of dependencies: Hard no from me, dude. 'No' means it'll be harder for us to develop, but I believe allowing plugin developers and operators alike to reason about their plugin as a single thing - even if it's packaged and distributed as many artifacts - is going to be important. Example:

`netflix/somedep` has Clouddriver, Orca, and Deck components. If we were building this plugin with Gradle, I'd expect 3 artifacts: `netflix/somedep-clouddriver-1.0`, `netflix/somedep-orca-1.0` and `netflix/somedep-deck-1.0`. When a clouddriver is configured to use `netflix/somedep` the download resolver would know to look for a `netflix/somedep-clouddriver` artifact. This isn't too far off from how Spinnaker service modules are written and published today.

Be interested for @christopherthielen to chime in on how Deck would need multiple artifacts (as you point out). Thoughts?

link108: the registry/ repository distinction makes sense 👍 

Ah, I see, I was missing the part about Kork configuring repositories to look at, but that makes total sense and is a common pattern 👍 

I'm happy with that tradeoff around distinction of dependencies, one of the main points of this project is to lower the barrier of entry and requiring plugin creators to know which service needs which dependency is asking a lot. I was thinking about lean binaries, only loading what you need, but I'd prefer to optimize for plugin developer experience.

rfc/plugins.md
robzienert: I added a bit about checkums. I don't intend to go deep on the artifact store or the supported protocols. I think investigating the possibility of using `kork-artifacts` might be a decent idea. We definitely don't want to reinvent the artifact repository story - for example, we would likely use Artifactory here at Netflix.

rfc/plugins.md
robzienert: I got rid of this section. 👍

rfc/plugins.md
robzienert: The exact shape of the protocol is not yet known.

rfc/plugins.md
```diff
@@ -0,0 +1,338 @@
...
+}
+```
+
+It is highly recommended to read the documentation of PF4J to fully understand the scope of this proposal: This proposal will not go in depth on PF4J internals or taxonomy.
+
+### Front50 as the Source of Truth
+Operators will need the ability to understand, at a glance, what plugins are installed and enabled or disabled, their versions, as well as any other metadata about a plugin. 
+Whether plugins are implemented as in-process or not, Front50 should be used to track and expose a single source of truth of what plugins are installed in any particular service, as well as which ones are enabled or disabled.
+
+The Front50 plugin registry will also be exposed through Gate, and can be consumed by Deck to load the correct plugins on application startup.
```
luispollo: This initially sounded to me like we would depend on Deck to load plugins for other parts of the (Spinnaker) application. Maybe replace "the correct plugins" with "UI plugins"?

rfc/plugins.md
```diff
@@ -0,0 +1,338 @@
...
+
+Plugin resources will be placed in the following locations:
+
+- Java services: `/opt/spinnaker/plugins/resources`
+- Deck: `/opt/deck/html/plugins`
+
+Plugins are placed in a different location for Deck so that it can serve plugin resources.
+
+#### VM Based Environment
+Plugins and their configurations are baked into image.
```
luispollo: Doesn't this directly contradict the paragraph on lines 245-248? If they're dowloadable at runtime, by definition they do not need to be baked into any images, no?

robzienert: They do not need to be, no, but they can be. I'll update to be a bit more clear.

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+
+```proto
+syntax = "proto3";
+
+// PluginManifest is used during registration of a plugin
+message PluginManifest {
+	// The name of the plugin, which is used as a fully 
+	// qualified identifier, including a namespace and 
+	// plugin name, ex: "netflix/fast-properties"
+	string name = 1;
```
link108: thoughts on making namespace be its own attribute? They are different attributes, so would be good to keep them separate. We've seen some issues when referencing plugin names via command line, see: https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1571344728019300?thread_ts=1571286805.015200&cid=CK9FK4XDF
the issue above is that the slash is being URL encoded, which spring does not like to have in a request path. Modifying Halyard to send plugin names (including slashes) via query param works, but that does not seem like good API design

robzienert: Sure, that sounds fine to me. cc @jonsie

robzienert: cc @jonsie

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+
+	// A short description of the plugin
+	// ex: "Provides rollout safety of runtime application 
+	// config via pipelines"
+	string description = 2;
+
+	// The name of the plugin author, ex: "Netflix"
+	string provider_name = 3;
```
jonsie: Should we call this `author_name` instead?  What exactly does `provider` refer to here?  I know we build Extension points, and those are provided, but are plugins "provided"?

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+Halyard configs can be used to inform Spinnaker of desired state, but whatever state is defined in Front50 will be used as the canonical source for plugins. 
+When a service starts up, it will ask Front50 what plugins need to be installed and which of those need to be enabled or disabled. Halyard’s role will only go as far as updating Front50’s repository of plugin metadata.
+
+Using Front50 as a source of truth will require extension of PF4J.
+
+```proto
+syntax = "proto3";
+
+// PluginManifest is used during registration of a plugin
+message PluginManifest {
```
plumpy: Is this mostly just cached data? The only things I think you really need from a config point of view are:
1) some sort of plugin ID
2) how to find that plugin (artifact data or a remote endpoint)
3) whether or not that plugin is enabled

Everything else (author, description, version, compatibiluty) will be queried from the plugin at startup (either a manifest in a jar or via a `PluginInfo` RPC service) and pushed to front50?

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+### Up/Downgrading Plugins
+This section applies to federated development scenarios, where the operator is not necessarily the same as the plugin developers; where the deploy cadence of a plugin does not adhere to the deploy cadence of the core Spinnaker services.
+
+The upgrading (or downgrading, from now on broadly upgrading) plugin story is different based on the particular plugin runtime.
+
+#### In-process
+All in-process plugins should be downloadable. 
+When a service starts up, it will ask Front50 for a manifest of plugins, downloading all installed plugins that are not already present on the filesystem (in the case of baked-in plugins) to a locally-accessible directory. 
+At this point, PF4J will scan the directory, installing them into the process and enabling them based on the information pulled from Front50. 
+Should Front50 be unavailable, the service will fail to start.
```
plumpy: > Should Front50 be unavailable, the service will fail to start.

This is new territory for Spinnaker, AFAIK. Currently every service can start up on its own and just has reduced functionality until its dependencies can be contacted. This is especially helpful when you're boostrapping Spinnaker for the first time. Is this really something we're willing to give up?

As far as I can tell, the main point of having front50 as the central repository is that it means operators can see all the installed and enabled plugins. What if instead the services pushed this info into front50 at startup? This seems feasible to me, especially since we don't support changing your plugins at service runtime...

rfc/plugins.md
```diff
@@ -0,0 +1,343 @@
...
+Properties files configure the application and `manifest.yml` lets halyard know where plugin resources are located.
+`manifest.yml` can also override properties from properties file.
+PF4J will handle loading plugins into a Spinnaker service, while `manifest.yml` will inform which resources should be deployed to each service.
+
+### Dependencies
+A series of new kork modules will be provided for libraries and services alike to consume to expose new extension points. 
+Under the covers, this introduces PF4J as a new global service dependency.
+
+In-process plugins, by default, will have their own ClassLoader that is isolated from other Plugin ClassLoaders. 
+These Plugin ClassLoaders will be children of the root ClassLoader, meaning they cannot have dependencies that conflict with the root ClassLoader.
```
plumpy: > they cannot have dependencies that conflict with the root ClassLoader

Given the sheer quantity of jars used by Spinnaker, this is going to be nearly impossible. We can't even always get this right inside a given Spinnaker service; imagine how hard it's going to be for a plugin. You think you can just tell everyone to use the dependencies from `kork` but either of these very common scenarios will screw that up:
1) they're going to add dependencies of their own (since they'll need additional jars to add functionality) which will have transitive dependencies and, say, upgrade the version of Guava without realizing it
2) they build the plugin against Spinnaker 1.16.3 but try to run it in 1.17.0

I've worked on several products that used plugins (most recently IntelliJ) and I can tell you this is guaranteed to cause issues that are super hard to reason about (but are at least generally trivial to reproduce).

I think what you want to do is this:
1) extension points are built against a super minimal classpath... ideally just the JDK and some simple data classes. Maybe a library like `protobuf` that's specifically designed to have no other downstream dependencies.
2) plugins are built against this extension point but don't include those classes in their distribution jar
3) the root classloader for plugins has _only_ the jars from the extension point and its (minimal) dependencies

If that makes sense to you, I can try to add some documentation about this as a follow-up PR?

rfc/plugins.md
plumpy: Frankly, if we're going to do this, why not just throw away halyard and store all the config in `front50`

robzienert: Throwing away halyard is a great idea! 🤣

If, let's say, a cloud provider were a plugin, would we actually want to allow clouddriver to start? That's not reduced functionality, it's just straight up broken.

Would it be preferential from an operator perspective to install a plugin into each service, rather than services asking front50 what plugins a service needs to install?

rfc/plugins.md
robzienert: Yes that makes sense. In fact that's what we were landing on (see the latest dependency cleanup PR I have in kork, and will follow-up with cleaning up orca-api as well).

It's also why we want to standardize around protobuf as the contract, to help avoid this class of problems - definitely agree it'll be a headache.

Would love if you could follow-up PR on this, thank you!

rfc/plugins.md
robzienert: Yes, this is mostly cached data. You're right that we won't need all of it and that most could/would be sourced from the plugins themselves.

I was speculating that this info would be valuable metadata for operators.

rfc/plugins.md
plumpy: Ha, yes, I think we're generally in agreement there.

It seems like the three pieces of config for each plugin are (id+plugin_location+enabled). I think you're proposing a model where that information is stored in front50. It's certainly not a terrible idea on its face, but its definitely different from the status quo, where all config is stored in yaml files within each service. I'm not convinced that plugin config is so fundamentally different from other types of config that it warrants splitting the config in two.

But more importantly (and less flippantly) I just feel like moving configuration data into front50 is a pretty big change that should probably happen in a separate discussion... there are likely other things we'd consider moving and other solutions we might want to explore (Spring Cloud Config Server?).

robzienert: On the front50 hard dependency issue: What if we treated the config similar to Spring PropertySources, where if front50 is unavailable, it would fallback to config shipped with each individual service (and subsequently, services would report in what plugins + versions they're running)? This would allow us to ensure that critical plugin functionality would always be present even if front50 is unavailable, but still be able to allow plugin developers to iterate and release plugins without requiring operators to manage them.

One of the use cases that we're trying to address is that plugins are not necessarily owned by the people who are operating Spinnaker. If config is forced to live in each service, we as operators would need to be involved in any plugin changes, which would pretty well defeat a large purpose of the plugin work.

Spring Config Server is a potentially viable solution, but you're right, it would require additional thought.

The key takeaway is that we as Spinnaker operators should not be required to be in the game of managing plugin configuration. It's entirely possible that many Spinnaker deployments would have the operators and plugin developers be the same people, but in a large organization, this would not be the case.

plumpy: But the operator still has to restart any service when a new plugin is added, a plugin is removed, or a plugin version is changed, right?

> There are no current plans for supporting a world where in-process plugins can be upgraded at runtime.

I don't get how you achieve an operator-free upgrade scenario when that is the case... What use cases are you addressing here?

robzienert: We were planning to have Spinnaker redeploy itself (rolling push? not sure) when a new plugin is either installed or updated.

plumpy: I guess I don't really understand the difference, then. With front50, you push the plugin info to `front50` then restart Spinnaker. With a yaml config, you rebuild the Spinnaker container with new YAML then restart Spinnaker with that new container. I guess you don't want devs to rebuild Spinnaker containers?

Can't you just use a `latest` tag in the plugin config (e.g. `netflix.com:clouddriver-nflx:latest`) and then you don't need to do a config change for an upgrade anyway?

plumpy: The new-install story would still involve a yaml change, but do you really want people flinging new plugins into Spinnaker without any awareness from the ops team? Seems weird to me, but maybe you do...

ajordens: Personally I like idea of resolvers and letting an operator opt-in/out of the `front50` dependency (or some other plugin registry).

We (Netflix) have some flexibility around build/bake time configuration, instance launch and process launch ... I'm not sure I know enough right now to commit to one direction.

I'd like to see us move forward with some experiments, I don't see any decisions at this stage being irreversible.

rfc/plugins.md
ajordens: This feels like the thing to get (more) correct initially (vs. the earlier discussion around plugin registries).

@plumpy's suggestion makes sense to me.

rfc/plugins.md
ajordens: Sure, total agreement with that.

A resolver that knew how to source from `front50` would be configurable, just as one that could read from local config or an s3 bucket, etc.

rfc/plugins.md
bpowell: One of the things @christopherthielen brought up in the UI design doc is SRI (https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). It would be nice to have this included for Deck resources to validate what the user is getting is correct.

rfc/plugins.md
robzienert: :100: I'm working on refactoring the code to have resolvers, the first and only implementation will source all plugin config from the Spring `Environment` PropertySource.

PR #44, opened 2019-10-10
feat(github): Adding membership.yml manifest

robzienert:
Current tracking of reviewers & approvers is spread in a bunch of places.
The desire is to move this into automation sourced from a file in the community repo.
This change does not include any automation, just provides an initial step.

The format on this file may change in the future, it's not prescribing anything.
Upon merging, the `reviewers.md` and `approvers.md` files in `spinnaker/spinnaker` will be deleted.

This also is another step towards cleanup of GitHub groups after splitting from Netflix ownership.

### DIFF
--- a/membership.yml
+++ b/membership.yml
@@ -84,10 +84,12 @@ organization:
     release-managers:
       description: People who are allowed to merge to release branches
       members:
+      - duftler
       - ezimanyi
       - louisjimenez
       - maggieneterval
       - plumpy
+      - ttomsu
     sc:
       description: Spinnaker Steering Committee
       members:
--- a/membership.yml
+++ b/membership.yml
@@ -1,0 +1,105 @@
+organization:
+  groups:
+    amazon:
+      description: Team for Amazon employees, so they can triage & manage their GitHub Project
+      members:
+      - allisaurus
+      - anshrma
+      - clareliguori
+      - edwinavalos
+      - pauljrob
+    docs-approvers:
+      description: Approvers for documentation repos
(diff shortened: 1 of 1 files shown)


plumpy:
How did you generate this? Can you check in the tool? Even if it's crappy one-off Perl or something, it seems better to have _something_, otherwise this seems destined to get out of date...

plumpy:
Oh, I see, this is more documentation, not actually tied in any way to the way the teams are structured. Fair enough.

plumpy:
Why is netflix-new-hires in there, then? Or amazon? I don't really see what purpose those serve...

robzienert:
This was crafted with love, no tooling. :(

I'd like this to _lead into_ tooling, so I took a shot at what groups we'd want initially. The Amazon one is needed for their triage permissions of their Project board, the netflix-new-hires is there so they can be tagged as reviewers more readily: I'd be happy to change the netflix-new-hires group to something else and more common, but that's its goal.

Whatever groups exist, I want the groups and their membership to be documented as code, which we can then get tooling around to enforce.

plumpy:
Okay, that's fair, I guess I just saw this and immediately thought: "oh yeah, I need to add `duftler` and `ttomsu` into `release-managers`" and almost went ahead and did that which would have made this thing out of date before you even submitted it.

But whatever, it's not the end of the world if people forget to update it, I suppose. It's no worse than what we have now, really.

robzienert:
I added them as release managers, so at least that's sorted...

PR #52, opened 2019-10-22
docs(governance): Update Approvers qualifications

robzienert:
There was no explicit language that someone needs to be a Reviewer first. Also, there was no language around the process of moving from Reviewer to Approver, so I added some mentioning that there is no arbitrary timeline for moving between roles.

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -50,12 +50,13 @@ Qualification guidelines
 * Is already fulfilling the Reviewer role
 * Belongs to an organization with significant stake in project
 * Known active participant in the community
+* Has reviewed 5 PRs of significant scope
 * Has submitted at least 5 PRs of significant scope that are merged
 * Sponsored by 2 Approvers
 
 [Current Approvers](https://github.com/spinnaker/spinnaker/blob/master/approvers.md)
 
-If you’d like to nominate yourself as an Approver, please create an [Issue in the community repository](https://github.com/spinnaker/community/issues). If you are already a Reviewer who meets the above Qualifications, you do not need to wait before submitting for Approver status.
+If you’d like to nominate yourself as an Approver, please create an [Issue in the community repository](https://github.com/spinnaker/community/issues). 
 
 ### SIG Leads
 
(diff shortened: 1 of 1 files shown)


duftler:
+1

governance.md
```diff
@@ -47,15 +47,15 @@ Duties
...
 * Belongs to an organization with significant stake in project
 * Known active participant in the community
 * Has submitted at least 5 PRs of significant scope that are merged
-* Has reviewed 5 PRs of significant scope
 * Sponsored by 2 Approvers
 
 [Current Approvers](https://github.com/spinnaker/spinnaker/blob/master/approvers.md)
 
-If you’d like to nominate yourself as an Approver, please create an [Issue in the community repository](https://github.com/spinnaker/community/issues). 
+If you’d like to nominate yourself as an Approver, please create an [Issue in the community repository](https://github.com/spinnaker/community/issues). If you are already a Reviewer who meets the above Qualifications, you do not need to wait before submitting for Approver status.
```
ezimanyi: I'm a bit confused by the new language.  So if someone belongs to an organization with significant stake in the project and wants to become an approver, they can:
* Submit an application for reviewer
* Immediately upon approval of that reviewer nomination, submit a new application for approver

I'm not sure I see the benefit in requiring someone to first become a reviewer if we're explicitly allowing them to immediately jump to approver (and dropping the requirement of having performed any reviewer duties).

ethanfrogers: Yeah, it does seem a tad confusing. Really, I think dropping the `Has reviewed 5 PRs of significant scope` back in would clear it up. If anything, it at least give you a bar to hit, since there isn't one defined in the `Reviewer guidelines` section.

demobox: +1 to having at least *some* additional criterion reviewers need to fulfil before being eligible to become approvers. Or do we genuinely feel there is no such criterion?

pstout: I agree with Andrew's comment.

robzienert:
Reverted most of the changes, only keeping the language that holding the reviewer status is required to become an approver.

demobox:
I'm fine with the change as proposed; just to clarify, though: there are no *additional* qualifications required for a Reviewer to become an Approver?

robzienert:
@demobox Currently, not really. I've found it difficult to find additional, objective qualifications that do not already get satisfied via Reviewer. Of course, I'd love to find some more and would be open to suggestions.

PR #62, opened 2019-11-12
chore(java11): Java 11 Migration RFC

plumpy:


rfc/java11.md
```diff
@@ -0,0 +1,164 @@
...
+JVMs have shown hesitancy to upgrade until the new runtime has undergone
+extensive testing. (Both Google and Netflix are in this category.)
+
+For official Spinnaker containers, we control the JRE used to run Spinnaker. For
+those releases, this proposal takes a relatively slow ramp to the Java 11 JRE
+and provides an escape hatch back to Java 8 for users who experience problems.
+
+There are also some unknown number of customers who aren't using the Spinnaker-
+provided containers. We will need to do some outreach to make sure these users
+convert to a Java 11 JRE too.
```
ethanfrogers: Do we have an idea of who these customers are? Also, do you have a plan for how this outreach will happen?

rfc/java11.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Timeline
+
+### 1.18 (early January)
+
+For our container builds, we'll first move just two (of approximately ten)
+services to the new JRE. Thee two services will be `igor` and `orca`. By
```
ethanfrogers: Out of curiosity, how were these services chosen?

rfc/java11.md
```diff
@@ -0,0 +1,164 @@
...
+discuss their concerns, which will allow us to adjust the timeline as needed.
+
+Separately, postings to `#dev` will encourage developers to start running their
+development environments under Java 11. Explanations of how to do so will be
+posted to the channel.
+
+If issues are found in Spinnaker that are caused by running under a Java 11
+runtime, we will switch the containers back to Java 8 and shift the timeline
+back one release. Since we're providing the `java8` variants, we will not adjust
+the timeline for any other reason.
```
ethanfrogers: Good call out! +1

rfc/java11.md
```diff
@@ -0,0 +1,164 @@
...
+
+For this release, we will do two things:
+
+1.  Compile with JDK 11 using the `-source 8 -target 8` flags. This produces
+    builds that are still compatible with Java 8, and means the build will still
+    fail if anyone tries to use Java 11 APIs or language features.
+2.  Remove the `java8` container variants
+
+If the compiler change causes issues, we can revert that change and push that
+out to the next release.
```
ethanfrogers: How does this relate to the above statement of `we will not adjust the timeline for any other reason`? I assume it means that we won't adjust the _JRE_ timeline for any other reason and that the _JDK_ timeline is separate.

rfc/java11.md
```diff
@@ -0,0 +1,164 @@
...
+2.  use the new JDK's `javac` with `-target 8 -source 8`
+3.  remove those flags
+
+Between each step is a period of time to evaluate breakages and potentially roll
+back.
+
+## Known Unknowns
+
+On the JRE side, there are probably changes to performance, memory usage, and
+anything else that could be affected by a major runtime change like this.
```
ethanfrogers: Do you have any plans for how to measure this other than users reporting perf issues? Would it be helpful to take a baseline against, say, the integration tests, and report any noticeable difference as the timeline progresses? This might give us an opportunity to set some users at ease when reading release notes by saying something like "In our tests, memory and CPU consumption was N% above/below our baseline". Also, if we see anything majorly concerning, we can address it prior to a release.

rfc/java11.md
```diff
@@ -0,0 +1,164 @@
...
+assumption is that companies that aren't using the container builds also have
+the in-house Spinnaker and Java expertise to handle this runtime change without
+too many problems.
+
+## Future Possibilities
+
+By the time this RFC is finished, Java 14 will have been released. This is the
+next LTS release of Java. This RFC should be updated with any changes to the
+plan so that it can be used as a template for the Java 14 upgrade, whenever we
+decide that should happen.
```
ethanfrogers: Are we hoping that, going forward, we continue to track more closely with LTS as they're release or handle this on a case-by-case basis?

rfc/java11.md
plumpy: I think they're pretty much by definition going to be customers who are very familiar with Spinnaker, since they're doing their own builds and deployments without the benefit of Halyard. So I don't imagine there are really any that we don't already have contact with...

In addition to the release notes, announcement emails, and slack updates, I figured I would get a list of names from our team and yours and reach out directly (via email) to the ones we can think of...

rfc/java11.md
plumpy: Somewhat arbitrarily. My criteria:
1) I wanted 2–3 services to start
2) I wanted some small and some a little "meatier"
3) but since the GC is the main thing I'm worried about (mostly because that's what I've heard Google is most concerned about), I wanted to avoid `clouddriver` initially, which I imagine will be the most sensitive to GC changes

Happy to discuss picking other services if you think there are better candidates...

ethanfrogers: I don't have any specific suggestions, I was just curious. +1 to having small/large. I'm a _touch_ worried about Orca because of how critical it is but I trust that the Platform SIG (I assume this falls under their realm) has thought through that.

ethanfrogers: I don't have any specific suggestions, I was just curious. +1 to having small/large. I'm a _touch_ worried about Orca because of how critical it is but I trust that the Platform SIG (I assume this falls under their realm) has thought through that.

rfc/java11.md
plumpy: I think we have to handle this case-by-case. We can't introduce a dependency on Java 14 until Netflix is comfortable running Java 14, for example. So it's going to have to be dependent on our customers being ready to run the new JREs.

That said, the LTS releases should be closer together now, so the diff between Java 11→14 is much smaller than 8→11.

rfc/java11.md
plumpy: I think instrumenting the integration tests in a way that would give us that data would be a very significant amount of work. We can't really just use a wall clock because the random latency from talking to cloud providers is almost certainly going to hide any changes from the JRE. So we'd have to stick a profiling tool in there somewhere and measure CPU usage and frankly I'm not sure the workloads are similar enough to give any useful data anyway. So I don't think the return on the investment would really be worth it.

rfc/java11.md
plumpy: The platform SIG hasn't discussed this yet, but I'm going to bring it up at our next meeting :grin:

rfc/java11.md
plumpy: I didn't really mean for that line to apply here. I pushed a commit that tries to clarify it more, let me know if it makes sense...

rfc/java11.md
```diff
@@ -0,0 +1,171 @@
...
+There are also some unknown number of customers who aren't using the Spinnaker-
+provided containers. We will need to do some outreach to make sure these users
+convert to a Java 11 JRE too.
+
+## Timeline
+
+### 1.18 (early January)
+
+For our container builds, we'll first move just two (of approximately ten)
+services to the new JRE. Thee two services will be `igor` and `orca`. By
```
robzienert:
```suggestion
services to the new JRE. The two services will be `igor` and `orca`. By
```

rfc/java11.md
robzienert: I'm not particularly concerned with selecting Orca, but if we want a "larger" service that iterates with less cadence, Front50 may be a good choice.

rfc/java11.md
```diff
@@ -0,0 +1,171 @@
...
+before, the `java8` variants will offer a way to opt out of this change.
+
+Other than that, everything from the previous section applies to this release as
+well.
+
+### 1.20 (early May)
+
+For this release, we will do two things:
+
+1.  Remove the `java8` container variants.
```
robzienert: I'm 95% sure Netflix is running a slower migration than Google is to Java 11. I haven't been given a solid timeline yet, only that it's in 2020 sometime.

rfc/java11.md
```diff
@@ -0,0 +1,171 @@
...
+those releases, this proposal takes a relatively slow ramp to the Java 11 JRE
+and provides an escape hatch back to Java 8 for users who experience problems.
+
+There are also some unknown number of customers who aren't using the Spinnaker-
+provided containers. We will need to do some outreach to make sure these users
+convert to a Java 11 JRE too.
+
+## Timeline
+
+### 1.18 (early January)
```
ttomsu: Seems like we could make these changes in the nightly build (i.e. now) and monitor changes prior to the 1.18 release.

ttomsu: Or were you intending to do this post-1.18 for inclusion in the 1.19 release in March?

rfc/java11.md
```diff
@@ -0,0 +1,171 @@
...
+
+The release notes will prominently mention this fact. It will also be mentioned
+in-channel in the release announcement to `#spinnaker-releases` and in the
+release announcement email. We will strongly encourage people who opt for the
+`java8` variants to contact `sig-platform@spinnaker.io` or `#sig-platform` to
+discuss their concerns, which will allow us to adjust the timeline as needed.
+
+Separately, postings to `#dev` will encourage developers to start running their
+development environments under Java 11. Explanations of how to do so will be
+posted to the channel.
```
ttomsu: s/channel/spinnaker.io website. Channel instructions scroll off the screen pretty easily.

rfc/java11.md
```diff
@@ -0,0 +1,171 @@
...
+
+Other than that, everything from the previous section applies to this release as
+well.
+
+### 1.20 (early May)
+
+For this release, we will do two things:
+
+1.  Remove the `java8` container variants.
+2.  Compile with JDK 11 using the `-source 8 -target 8` flags. This produces
```
ttomsu: There are various places throughout the Gradle plugins that specify/override these settings, so coordination with Netflix's gradle plugins team may be necessary. @cfieber probably knows more about this.

rfc/java11.md
plumpy: Good call, thanks I added this:

> The [Spinnaker Developer Guide](https://www.spinnaker.io/guides/developer/getting-set-up/) will also be updated to strongly recommend using Java 11.

rfc/java11.md
plumpy: FYI, my plan is to do them as soon as this RFC is accepted. But I'm really only documenting the user-visible changes, not changes to the build process that only really affect us... (I do mention this point in non-goals.) We should definitely talk internally and make sure we're on the same page about build system changes, though...

rfc/java11.md
plumpy: Yep, definitely, but these details are also non-goals of this doc...

> Non-goals:
> * Documenting and solving technical issues that present themselves in the migration; those will be handled separately, but this document concerns itself primarily with the larger processes and timeline

rfc/java11.md
plumpy: My inclination is to start migrating the containers anyway and obviously hold off on the final switch to `-source 11 target 11` until you're ready (and everyone else is too). Having other people move to Java 11 feels like it could only boost your confidence in the move, so would still be a win for you. But happy to discuss holding off on this if you disagree.

Obviously it isn't super urgent to do this, and holding off a little longer isn't a huge deal. I just wanted to get the process started since it's a relatively a long timeline...

rfc/java11.md
```diff
@@ -0,0 +1,173 @@
...
+*   Discuss ways we will encourage people using non-offical builds to also move
+    to the Java 11 JRE
+*   Provide a timeline for the final switch to a Java 11 JDK along with criteria
+    to decide if the timeline should be delayed
+
+Non-Goals:
+
+*   Providing a rationale for upgrading Java; all active Java projects need to
+    update their JDK version eventually as libraries we depend upon will also
+    depend on new versions of the JDK
```
seanpeters86: I think this is a good point, ideally everything would upgrade at the same time (same release window) as maintaining multiple services under different versions would not be ideal.

rfc/java11.md
seanpeters86: We use some "custom" builds but for the most part these are built off of the current builds that halyard pulls from anyways. So the upgrade process would be similar for us anyways.

rfc/java11.md
```diff
@@ -0,0 +1,173 @@
...
+There are also some unknown number of customers who aren't using the Spinnaker-
+provided containers. We will need to do some outreach to make sure these users
+convert to a Java 11 JRE too.
+
+## Timeline
+
+### 1.18 (early January)
+
+For our container builds, we'll first move just two (of approximately ten)
+services to the new JRE. The two services will be `igor` and `orca`. By
```
seanpeters86: Any reason you picked these two? Igor I can see as not "everyone" will be using it, but orca is pretty core to spinnaker and typically something everyone is running (even if they aren't running every microservice). Maybe `igor` and `echo` would be the easiest?

rfc/java11.md
```diff
@@ -0,0 +1,173 @@
...
+default, anyone who uses Halyard to deploy these containers will get the Java 11
+JRE for these services. Users who experience problems or aren't ready to use
+Java 11 can use the `java8` variant by specifying `imageVariant: slim-java8` (or
```
seanpeters86: Does this give the community much benefit? Depending on how much work it is to maintain both versions this might be more harm than anything. If people want Java 8 versions they could just run 1.17 or earlier versions until they are ready to upgrade. Making sure every commit is compatible with both versions might be a hassle. Might be easiest to just rip the band aid at the next release to Java 11.

rfc/java11.md
seanpeters86: I would just add that upgrading for the sake of upgrading is probably not worth the risk unless we are getting something of benefit from doing so

plumpy: I generally agree with that, although we do have to upgrade at some point just to keep up with the Java ecosystem, regardless of whether or not we directly are benefiting. At some point libraries we depend on will stop supporting old JREs because they're compiled with a newer JDK, etc. Upgrading too early has its own risks, but I'd also personally prefer to be a little proactive so we don't end up in a situation where our hand is forced because we need to upgrade some important library and can't until we go through the six month JDK upgrade process detailed here...

rfc/java11.md
plumpy: I guess it disappeared from the diff view because I pushed a new commit, but there was more discussion about this here: https://github.com/spinnaker/community/pull/62#discussion_r345726643

(The services chosen are definitely very much up for discussion.)

rfc/java11.md
plumpy: Since the builds are still going to be compiled with a Java 8 JDK, there isn't really any worry about compatibility from a developer perspective. If you try to use a Java 11 language feature or library, the compilation will fail, hard stop. You have to keep writing as though we were still using Java 8.

The band-aid rip will only be when we switch to a Java 11 JDK and use the flags `javac -source 11 -target 11`. Only when we do that will developers be able to use new Java 11isms, and we will do that for all repositories at once.

This is really just about which runtime environment (`java` binary) is used.

rfc/java11.md
plumpy: But are you building your own containers? The important thing here is that those containers would need to be switched to a Java 11 JRE, if they aren't already. I guess if you use the bundled Dockerfiles, this would be handled for you. (I'll just switch the first line of this file to openjdk 11: https://github.com/spinnaker/orca/blob/master/Dockerfile.slim)

rfc/java11.md
ttomsu: > That really only affect us

 Not for long!

rfc/java11.md
cfieber: We also have `kork` (significantly) and `fiat` (its client lib) exporting dependencies into other projects. There might be a step 0 here that ensures that both of those are compatible with both JRE 8 and JRE 11.

rfc/java11.md
robzienert: I agree - I'm playing with the idea of adopting the alpha java11 base image here at least for some of our preprod environments in Q1. I don't need java11 marked as stable to feel confident in getting Spinnaker prod out there as long as it looks generally cool under some load.

rfc/java11.md
seanpeters86: Yes we are building our own containers (to build in our own internal stuff into the services), but these are just references of the existing images that are published, so not a big deal for us.

rfc/java11.md
plumpy: Switched `orca` to `front50`.

plumpy:
Discussed this in the platform SIG and the TOC. No real objections, so I'm going to merge this.

PR #68, opened 2019-11-17
fix(links): point to correct file and line for reviewers/approvers

xavileon:


governance.md
```diff
@@ -33,7 +33,7 @@ Qualification guidelines
 * Has submitted at least 5 PRs of significant scope that are correspondingly merged
 * Sponsored by 2 Approvers
 
-[Current Reviewers](https://github.com/spinnaker/spinnaker/blob/master/reviewers.md)
+[Current Reviewers](https://github.com/spinnaker/community/blob/master/membership.yml#L23)
```
demobox: @dorbin What would be an appropriate link here? Linking to particular lines in a file could be brittle; is the membership "rendered" anywhere with a potentially more stable link?

dorbin: How about just linking to the file? The sections within it are pretty clear, and it's not that huge.

governance.md
```diff
@@ -54,7 +54,7 @@ Qualification guidelines
 * Has reviewed 5 PRs of significant scope
 * Sponsored by 2 Approvers
 
-[Current Approvers](https://github.com/spinnaker/spinnaker/blob/master/approvers.md)
+[Current Approvers](https://github.com/spinnaker/community/blob/master/membership.yml)
```
demobox: @dorbin It _seems_ reasonably clear that "oss-reviewers" and "oss-approvers" are the sections readers should find, but there's e.g. also a "oss-kayenta-approvers" section that could cause confusion.

Perhaps something like `[Current Approves](...) (see oss-approvers)` or so?

dorbin: Yes, like that. But "...(see the `oss-approvers` section)."

Also, prefer lower-case "approvers" and "reviewers".

demobox: > Also, prefer lower-case "approvers" and "reviewers".

"Approvers" seems capitalized elsewhere in the doc too, and appears to be being regarded as a specific title?

demobox:
```suggestion
[Current Approvers](https://github.com/spinnaker/community/blob/master/membership.yml) (see the `oss-approvers` section)
```

governance.md
demobox:
```suggestion
[Current Reviewers](https://github.com/spinnaker/community/blob/master/membership.yml#L23) (see the `oss-reviewers` section
```

PR #74, opened 2019-12-06
feat(rfc): plugins ui rfc

bpowell:
Migrating this RFC from https://docs.google.com/document/d/16WmRSziTJsSBZ1kuKUfVleLAYMIxmfvz/edit

### DIFF
--- a/rfc/plugins-ui.md
+++ b/rfc/plugins-ui.md
@@ -1,0 +1,111 @@
+# Plugins UI Design Doc
+## Motivation
+Just like the backend in Spinnaker, the frontend needs to be extensible for others to add more functionality. Deck should be able to be extended by well defined extension points. Some examples of extension points could be:
+1. Custom stages
+2. Modifying header and footer
+3. Custom search filters
+4. Custom routes
+5. Adding labels or other information to sections in the infrastructure tab
+
+This design doc references the [Plugins RFC](./plugins.md).
+
+This doc focuses on
+* How plugins are implemented in Deck
+* Deck  Plugin Interaction
+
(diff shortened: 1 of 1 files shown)


rfc/plugins-ui.md
```diff
@@ -28,24 +28,20 @@ In order for plugins to have a UI component, they must be loaded into Deck’s r
...
 Please see the current [Plugins RFC](./plugins.md) for details around Front50 as the source of truth for plugin metadata. 
 
-
 ### Gathering plugin metadata
-The Deck microservice does not need to download plugin resources; however, the browser does need to know where the plugin resources are located. 
-
-Deck provides a function to lookup plugin metadata in Front50, which is exposed via Gate, so the browser can gather that information. In order for resources to be loaded appropriately, Deck must gather the resources prior to the application being bootstrapped.  
+Using native Javascript `import`, plugin resources have to be served up by Deck. `import` does not allow resources to be imported from external sources. Therefore all plugin resources need to be added to Deck before Deck starts.
 
+In a Kubernetes world, Halyard can place the plugin resources into an init container for Deck to consume. The init container will download the plugin resources and place the plugins into their own folders. The plugins will be mounted into the Deck container under `/opt/deck/html/plugins/`.
```
ezimanyi: Could you elaborate a bit on what the process would be for putting these resources into an init container?  I'd like to avoid putting more service-specific logic into Halyard, which is already quite complex and difficult to maintain.  If this described in more detail how that init container would be created, we could figure out where that logic would best live.

robzienert:
TOC +1. Spirit of the direction is there and we're already well into implementing it.

PR #82, opened 2019-12-16
chore(sigs): add sig calendar

brian-armory:
I'm adding the calendar info to the sig-index page since we're using that as the source of truth.

When the governance info was removed from spinnaker.io, the calendar info for SIGs disappeared with it (https://github.com/spinnaker/spinnaker.github.io/pull/1598).

### DIFF
--- a/sig-index.md
+++ b/sig-index.md
@@ -11,3 +11,17 @@ Don't see a SIG here that fits what you're looking for? [Submit a Proposal](sig-
 * [Spinnaker-as-Code](sig-spinnaker-as-code/README.md)
 * [Summit](sig-summit/README.md)
 * [UI / UX](sig-ui-ux/README.md)
+
+# SIG Meeting Calendar
+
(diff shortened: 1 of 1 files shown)


sig-index.md
```diff
@@ -11,3 +11,17 @@ Don't see a SIG here that fits what you're looking for? [Submit a Proposal](sig-
 * [Spinnaker-as-Code](sig-spinnaker-as-code/README.md)
 * [Summit](sig-summit/README.md)
 * [UI / UX](sig-ui-ux/README.md)
+
+# SIG Meeting Calendar
+
+To view the calendar, [click here](https://calendar.google.com/calendar/b/3/embed?showPrint=0&showCalendars=0&mode=AGENDA&height=500&wkst=1&bgcolor=%23FFFFFF&src=spinnaker.io_tr65rjf5mfij7p6vucprkhulcc%40group.calendar.google.com&color=%2342104A&ctz=America%2FLos_Angeles).
```
ezimanyi: When I click this link, the calendar is titled "Michael Plump"...is there some way to change that?  Also, I wonder if we want to remove the `ctz=` field from the URL so that it defaults to the user's timezone and not always America/Los_Angeles.

brian-armory: I noticed the "Michael Plump" thing too. I pinged @plumpy about it in the sig-docs channel. 

I can remove the `ctz` part. It's just a carryover from what was on the .io page.

PR #85, opened 2019-12-18
docs(governance): the community repo was renamed to governance

plumpy:
I suppose the info from the README should move back to the website now?


robzienert:
I don't think we need to change the README?

PR #89, opened 2020-01-03
feat(rfc): RFC to replace Halyard

ezimanyi:
The current recommended way to deploy Spinnaker is by using Halyard. While Halyard has greatly simplified some aspects of installing and configuring Spinnaker, it has a number of shortcomings that are particularly apparent when deploying to Kubernetes. This document proposes replacing Halyard with a smaller tool that integrates better into the Kubernetes ecosystem.

### DIFF
--- a/rfc/halyard-lite.md
+++ b/rfc/halyard-lite.md
@@ -238,9 +238,13 @@ firewalls prevent the Halyard instance from communicating with external services
 individual microservices do have access to.
 
 Halyard will reduce the scope of its validation efforts.  It will focus on
-validating the well-formedness of values in the Halyard config, but will no longer
+validating the well-formedness of the Halyard config, but will no longer
 act as the other microservices to try to actually build instances of credentials
-and make requests using them.
+and make requests using them.  Validating the well-formedness of the config will
+involve both:
+* Validating the structure of the config
+* Performing simple validation on the values supplied in the config, such as
+  ensuring that a value is an integer or a valid URL
 
 #### Secret management
 
@@ -279,9 +283,17 @@ that the operator is using some external tool to version and back up the configu
 Halyard handles publishing and deprecating Spinnaker versions; this functionality
 is only used by maintainers of the project and requires having a GCP service
 account with write permissions to the GCS bucket that stores Spinnaker version
-info. To simplify the surface area for end users, this functionality will be
+info. It is possible to use this functionality to publish custom BOMs to a private
+GCS bucket, so users who publish and consume custom BOMs may also be using these
+commands.
+
+To simplify the surface area for end users, this functionality will be
 removed from Halyard and migrated to a separate admin-specific tool to handle
-these changes.
+these changes. Any users that are using the `hal admin` commands to publish
+custom BOMs will need to migrate to the new tool; given that users publishing a
+custom BOM are in general advanced users, this should not be a difficult migration.
+(More research will also be done at that stage to determine if there are even any
+end-users relying on these commands as part of their custom BOM workflow.)
 
 ## Retained core competency
 
--- a/rfc/halyard-lite.md
+++ b/rfc/halyard-lite.md
@@ -1,0 +1,473 @@
+# Replacement of Halyard
+
+| | |
+|-|-|
+| **Status**     | _**Proposed**, Accepted, Implemented, Obsolete_ |
+| **RFC #**      | [89](https://github.com/spinnaker/governance/pull/89) |
+| **Author(s)**  | Eric Zimanyi (@ezimanyi) |
+| **SIG / WG**   | Kubernetes SIG |
+
+## Overview
+
+The current recommended way to deploy Spinnaker is by using Halyard. While Halyard
+has greatly simplified some aspects of installing and configuring Spinnaker, it
+has a number of shortcomings that are particularly apparent when deploying to
+Kubernetes. This document proposes replacing Halyard with a smaller tool that
+integrates better into the Kubernetes ecosystem. Throughout this document, the new
+tool is referred to by the temporary name _Halyard Lite_.
+
+### Goals and Non-Goals
+
+Goals:
+* Improve the install process for users deploying to Kubernetes by providing an
+install path that is Kubernetes-native and integrates well with the Kubernetes
+ecosystem
+* Reduce the maintenance burden of Halyard, in particular by reducing the toil
+required to support new configuration parameters
+
+Non-goals:
+* This document focuses primarily on the motivation behind replacing Halyard with
+a smaller tool as well as the broad design and surface of of this tool. A more detailed
+technical design of this tool is not in scope for this document.
+
+## Motivation and Rationale
+
+The primary purpose of Halyard is to simplify the process of configuring, updating,
+and operating Spinnaker.
+
+Without Halyard, installing Spinnaker involves:
+*   Determining the version of each microservice you’d like to deploy.  In general,
+this would involve picking a top-level Spinnaker version (ex: 1.16.1) and looking
+up the version of each microservice tied to that top-level version.
+*   Writing a `service.yml` config file for each microservice containing the
+configuration you want for your Spinnaker installation, and copy it to the
+correct directory.
+    *   The configuration parameters accepted by each microservice are in general
+    not well documented, so writing these config files would require a fairly
+    detailed understanding of each microservice.
+
+Installing to a Kubernetes cluster has the additional complexity of:
+*   Adding each service’s config file as a `ConfigMap` to your cluster
+*   Writing a workload (`Deployment`/`ReplicaSet`) manifest for each microservice,
+making sure to mount the appropriate configuration into each workload.
+*   Writing a `Service` for any workloads (deck, gate) that should be exposed
+outside of the cluster.
+
+With Halyard, this process is greatly simplified.  In particular, a user can
+install Spinnaker to Kubernetes via:
+*   Configuring Halyard with the top-level Spinnaker version to install, without
+needing to know about the individual microservice versions
+*   Adding all configuration parameters to a single config file, either by editing
+the YAML directly or by running `hal config` commands
+*   Running `hal deploy apply` to deploy Spinnaker
+
+While Halyard has greatly simplified the install process, it has a number of shortcomings.
+
+### Not Kubernetes Native
+
+As Spinnaker positions itself as a Kubernetes-native deployment tool, operators
+are expecting to deploy and operate Spinnaker using the same Kubernetes-native
+tools they use for other software.  In particular, the workflow for updating
+Spinnaker using Halyard is to SSH to a VM (or a pod) and run a series of imperative
+commands.
+
+The desire for a Kubernetes-native path to install Spinnaker has led to a number
+of alternate install paths built on top of Halyard:
+*   A [Helm chart](https://github.com/helm/charts/tree/master/stable/spinnaker) that
+installs Halyard as a stateful set
+    *   While this has provided a Kubernetes-native way of installing Halyard,
+    the process of running and maintaining Spinnaker itself still involves SSH’ing
+    to the Halyard pod and running imperative commands
+*   An operator built by [OpsMx](https://operatorhub.io/operator/opsmx-spinnaker-operator)
+    *   This operator appears to have many of the same drawbacks as the Helm chart,
+    in that it will initially configure a Spinnaker installation but many customizations
+    still involve SSH’ing to a Halyard pod and running imperative commands.
+*   An operator built by [Armory](https://docs.armory.io/spinnaker/operator/)
+    *   This operator removes the need for users to run imperative commands to
+    manage Spinnaker.  It allows users to configure Spinnaker via CRUD on a
+    SpinnakerService CRD that contains the full Halyard configuration, with the
+    operator handling any redeployments from those changes.
+    *   While much of the surface area of Halyard is hidden from users of this
+    operator, it is still built on top of Halyard.
+
+### Duplication of Kubernetes-native tools
+
+Users often have requirements for customizing the Kubernetes objects that comprise
+a Spinnaker installation.  For example, a common use case is to add a toleration
+to some/all of the workloads so that they can run on specific node pools.
+
+As Halyard is responsible for both generating and deploying the YAML for all the
+Kubernetes objects that comprise a Spinnaker installation, Halyard needs to handle
+any customization that users want to apply.  Maintaining all of these possible
+customizations in Halyard adds a maintenance burden, as users frequently open pull
+requests to add additional fields they would like on their deployments/services.
+Changes to the templates that generate the YAML Halyard deploys are among the most
+frequent causes of bugs, as there is no test coverage on these files and the templating
+language (Jinja) is often not familiar to contributors.
+
+The problem of taking a set of YAML files that constitute a software installation,
+and customizing them with various overrides is one that has numerous solutions in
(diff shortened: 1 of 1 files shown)


rfc/halyard-lite.md
```diff
@@ -0,0 +1,473 @@
...
+
+While Halyard has greatly simplified the install process, it has a number of shortcomings.
+
+### Not Kubernetes Native
+
+As Spinnaker positions itself as a Kubernetes-native deployment tool, operators
+are expecting to deploy and operate Spinnaker using the same Kubernetes-native
+tools they use for other software.  In particular, the workflow for updating
+Spinnaker using Halyard is to SSH to a VM (or a pod) and run a series of imperative
+commands.
```
ethanfrogers: Adding a bit more color - the current implementation doesn't lend itself well to any style of GitOps or IAC (Infrastructure as Code) workflows because of Halyard's client/server nature. While are have seen a number of users work around this, the process for doing so is clumsy and error prone making it difficult to automate.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,473 @@
...
+  * Feedback from early users of Halyard Lite and the kustomization are
+   addressed, and the kustomization now supports more advanced use cases (ex:
+   high availability, custom BOM, etc.)
+  * The Armory operator is moved to depend on Halyard Lite
+* Spinnaker 1.21 (July 2020)
+  * Halyard Lite, as well as the kustomize install path are GA and the
+   recommended install path for new users of Spinnaker.
+  * Based on adoption of Halyard Lite, a plan is formulated around how and when
+  to deprecate and cease support of Halyard. The details of this plan will be a
+  separate RFC.
```
ethanfrogers: I assume this also includes plans for updating the documentation?

ezimanyi: Yes, good point!  There will of course be documentation on Halyard Lite accompanying the alpha release so that early adopters can use it---but as part of the deprecation of Halyard we'll need to update all of the documentation that references `hal` commands to instead show how to update the halyard config.  I'll update the RFC to make this explicit.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+#### CLI vs. Daemon
+
+Halyard is currently implemented as a CLI and a daemon, greatly adding to its
+complexity.  This decision was made to allow support for clients other than the
+CLI, but to date no such implementations have been written and none are planned.
+The new tool will replace this with a CLI only, with no background daemon process.
+
+#### Language
+
+Halyard Lite will be written in Go, which has the following advantages:
```
ttomsu: +100 to be written in Go - I would also strongly suggest code authors and reviewers be well versed in idiomatic Go, as it sometimes has quirks not present in other languages. (cc @brandonnelson3)

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+parameters.  (For example, when configuring a Google Compute account, Halyard will
+try to make a test request to list instances during its validation step.)
+
+The value of this validation is not clear when Halyard is running on a different
+machine than the microservice in question.  In many cases, network policies or
+firewalls prevent the Halyard instance from communicating with external services that
+individual microservices do have access to.
+
+Halyard will reduce the scope of its validation efforts.  It will focus on
+validating the well-formedness of values in the Halyard config, but will no longer
```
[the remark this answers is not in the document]
ezimanyi: I think we'd support some amount of active validation---here I'd include "well-formedness" to be more than just if something is a `string` or `int32` but also to include, for example, whether it is a valid URI.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+To avoid duplicating existing solutions, Halyard will no longer have built-in
+support for backing up and restoring the configuration; it will instead presume
+that the operator is using some external tool to version and back up the configuration.
+
+#### Version publishing/deprecating
+
+Halyard handles publishing and deprecating Spinnaker versions; this functionality
+is only used by maintainers of the project and requires having a GCP service
+account with write permissions to the GCS bucket that stores Spinnaker version
+info. To simplify the surface area for end users, this functionality will be
```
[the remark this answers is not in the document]
ezimanyi: Ah, good point---I hadn't realized that people might be using the `hal admin` commands for custom BOMs.  I *suspect* that they aren't and are just editing the BOMs we publish to change the container references, but will validate this with users.  Either way I don't think it substantially changes the RFC; we'll to continue to support a way to publish a BOM for our internal use and can transition users publishing custom BOMs over to that tool.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+Kubernetes-native.  The CLI also encourages an imperative workflow whereby
+configuration is achieved via a series hal commands.  Furthermore, a significant
+amount of the complexity and verbosity of Halyard’s codebase comes from the commands
+to edit the Halyard configuration.
+
+Based on these considerations, the CLI commands to edit the configuration will be
+removed.  As the [hal command reference](https://www.spinnaker.io/reference/halyard/commands/)
+is the best existing documentation on the available config parameters for Spinnaker,
+it will be replaced by extensive documentation of a well-typed Halyard configuration
+so that users can be confident directly editing this config.
```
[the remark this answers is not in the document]
ezimanyi: Yes, this is an important part of deploying to Kubernetes.  Halyard Lite will stage the files just as Halyard did, but it will then be the responsibility of Helm/Kustomize to create config maps/secrets out of those files and deploy them to the cluster.

I think this actually opens up for better Kubernetes-native solutions going forward...such as just directly referencing a secret in the config file and decoupling the deployment of Spinnaker from the deployment of the tokens that it uses.

rfc/halyard-lite.md
ezimanyi: Yes, we definitely plan on making sure to learn Go's idioms to avoid writing Java that has been lightly translated to Go.  We're also working with @ncknt who's written a lot of Go (including the Armory operator).

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+Kubernetes. This document proposes replacing Halyard with a smaller tool that
+integrates better into the Kubernetes ecosystem. Throughout this document, the new
+tool is referred to by the temporary name _Halyard Lite_.
+
+### Goals and Non-Goals
+
+Goals:
+* Improve the install process for users deploying to Kubernetes by providing an
+install path that is Kubernetes-native and integrates well with the Kubernetes
+ecosystem
```
robzienert: Interesting wording here. Today, Halyard is _not_ Kubernetes specific. It sounds like this proposal is saying that Kubernetes will be the only supported deployment path for Spinnaker through Halyard?

I understand there's Kubernetes-native, but have we figured out what the story will be for operators that use Halyard today, but not K8s?

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+Spinnaker using Halyard is to SSH to a VM (or a pod) and run a series of imperative
+commands.
+
+The desire for a Kubernetes-native path to install Spinnaker has led to a number
+of alternate install paths built on top of Halyard:
+*   A [Helm chart](https://github.com/helm/charts/tree/master/stable/spinnaker) that
+installs Halyard as a stateful set
+    *   While this has provided a Kubernetes-native way of installing Halyard,
+    the process of running and maintaining Spinnaker itself still involves SSH’ing
+    to the Halyard pod and running imperative commands
```
robzienert: Whoa! I had no idea that this was the process for Halyard users. Gross.

rfc/halyard-lite.md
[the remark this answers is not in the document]
robzienert: So really only validating rich types, in addition to structure? I'd be 💯 about this.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+In order to promote backwards compatibility, the format of this config file will
+not change.  There will be some fields that are no longer relevant in Halyard Lite;
+these fields will be documented as being ignored by Halyard Lite and it will emit
+a non-fatal warning if these fields are set.  (An example of such a field is
+`deploymentEnvironment.tolerations`, which does not affect the generation of
+service configurations but only adds fields to the output deployment YAML.)
+
+As the Halyard config is not currently well-documented (with the documentation
+instead living on the hal command reference), we will add extensive documentation
+on this config file.
```
robzienert: Would this documentation live within the code, or be separate? 

I have a strongly held opinion that it should live with the code and create doc artifacts during build. This is similar in direction to an upcoming backend config proposal I have coming through the pipes.

rfc/halyard-lite.md
robzienert: Following up on myself (and please correct if wrong): There's nothing in this proposal that makes Halyard necessarily Kubernetes-only, just that it's more heavily indexing on a smooth k8s experience.

rfc/halyard-lite.md
robzienert: :100: Another advantage is that you don't need to spin up the JVM to run some simple commands. That lag is infuriating for command-line tooling.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+
+## Known Unknowns
+
+At this point, there are still a few details that are unknown and will be clarified
+during the implementation and testing phases:
+* The exact format of the output from the CLI tool. It will need to output the service
+configs as well as (possibly) some metadata about the Spinnaker deployment (ex: the versions
+of containers to pull) but the exact format will be determined in concert with early consumers.
+* The exact format of the Helm chart and Kustomize kustomization that will be provided to support
+installation to Spinnaker remains unspecified.
```
robzienert: Also: How do we correctly communicate that, yes, Spinnaker is k8s native via this work, without making non-k8s users feel abandoned or pressured to move onto k8s.

rfc/halyard-lite.md
ezimanyi: This is a good point, and I'll update the RFC to be a bit more clear about it.  In general the primary motivation here is to make the experience easier for k8s users, because the coupling of config generation and deployment makes it particularly difficult for them to use tools they are used to using.

But as the output of Halyard Lite will just be your config files and some metadata about the service versions you want to deploy, there's nothing here that ties it specifically to Kubernetes.  Users deploying to VMs can use the output as part of their deployment process.

Halyard will no longer handle actually deploying these services to VMs (which it just did via some bash wrapping an `apt-get` with a pinned version), but given that it could only deploy all services to the same VM as Halyard it's not clear to me that this was overly useful beyond test setups anyway.  Just as with k8s, it will be up to downstream tools/scripts to actually deploy the services using the config generated by Halyard.

rfc/halyard-lite.md
ezimanyi: This would absolutely live alongside the code; the documentation will effectively just be documentation on the types that comprise the hal config.  I'm currently looking into the best way to auto-generate this among the various tools out there.

rfc/halyard-lite.md
ttomsu: One of my hopes with the stats collection is to be able to quantify these groups - how many deployments are going against which providers.

rfc/halyard-lite.md
ezimanyi: Well said, I've added this directly to the RFC as it provides useful context.

rfc/halyard-lite.md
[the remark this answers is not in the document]
ezimanyi: Added managing files to the list of things that were complex pre-Halyard as this is a good point.

rfc/halyard-lite.md
ezimanyi: Added this explicitly to the RFC.

rfc/halyard-lite.md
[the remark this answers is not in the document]
ezimanyi: Added this directly to the RFC.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,475 @@
...
+
+The problem of taking a set of YAML files that constitute a software installation,
+and customizing them with various overrides is one that has numerous solutions in
+the Kubernetes community, the most notable being [Helm](https://helm.sh/) and
+[Kustomize](https://github.com/kubernetes-sigs/kustomize).  Ideally, Halyard would
+not duplicate functionality that has existing solutions with much higher adoption;
+this would reduce both the maintenance burden on Halyard and the barrier to entry
+to Spinnaker for users familiar with these tools.
+
+### Dynamic configuration
```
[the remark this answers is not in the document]
ezimanyi: I think in general the approach is that config files can live in a source repository, and the running service can directly reference that repository.   In general, that works fine, as the config can be directly edited in the desired repo then be picked up by the service.

Where this doesn't interact well with Halyard is that it generates the service configs as transient files and then directly applies them; the generation of config is coupled to the deployment of that config.  After the changes in this RFC, I could imagine someone wanting to use Spring Cloud config deciding to commit the files generated by Halyard Lite to a config repo.  So the workflow would be:
* Edit the checked in hal config
* As part of the build step, run Halyard Lite to generate service configs, and also check those in
* Commit the full changes

Then when you deploy services, instead of deploying configs alongside them, just point them at your repo.

Let me know if this answers the question!

rfc/halyard-lite.md
robzienert: I'm not sure I follow - are you implying that we're hoping k8s has significant enough adoption we don't need to address the rest?

rfc/halyard-lite.md
ezimanyi: Added a section "Non-Kubernetes" to "Install Pathways"

rfc/halyard-lite.md
ezimanyi: I've added a point to "Non-goals" to explicitly state that this document does not intend to deprecate non-Kubernetes install pathways (even as it focuses on making Kubernetes installs easier).

rfc/halyard-lite.md
```diff
@@ -0,0 +1,508 @@
...
+parameters.  (For example, when configuring a Google Compute account, Halyard will
+try to make a test request to list instances during its validation step.)
+
+The value of this validation is not clear when Halyard is running on a different
+machine than the microservice in question.  In many cases, network policies or
+firewalls prevent the Halyard instance from communicating with external services that
+individual microservices do have access to.
+
+Halyard will reduce the scope of its validation efforts.  It will focus on
+validating the well-formedness of the Halyard config, but will no longer
```
ajordens: 👍

ajordens:
Love it, thanks for the detailed RFC @ezimanyi.

rfc/halyard-lite.md
```diff
@@ -0,0 +1,508 @@
...
+act as the other microservices to try to actually build instances of credentials
+and make requests using them.  Validating the well-formedness of the config will
+involve both:
+* Validating the structure of the config
+* Performing simple validation on the values supplied in the config, such as
```
duftler: Halyard also performed some additional simple validation that doesn't fall into the well-formedness or auth* categories: 
```
+ Get current deployment
  Success
+ Get persistent store
  Success
Validation in default.persistentStorage:
- WARNING Your deployment will most likely fail until you configure
  and enable a persistent store.
```

Will simple checks like this still be in scope?

ezimanyi: I think checks like that would be in scope; I suppose that's a bit broader than just checking each individual value, but I could imagine things of the type "you must have at least one of X" configured would be in scope.

PR #94, opened 2020-01-16
feat(rfc) RFC to enhance the Spinnaker Kubernetes UI

louisjimenez:
Proposal for changes to Spinnaker's Kubernetes UI.


rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+
+Similar to the effort to expose PVCs under a dedicated Storage tab, the UI components can be extended to surface Config Maps. These Config Maps will live in an additional UI tab for Configuration. A Config Maps’ stored data, a YAML viewer, and the following Metadata will be displayed:
+
+- Name
+- Namespace
+- Labels
+- Annotations
+- Created timestamp
+- Cluster
+
```
dbyron0: Some kind of connection between deployments (or other objects) and the configmaps they use would be helpful.  And if there's enough information available to indicate that a configmap isn't being used at all, that would be huge.  Even if cleanup is left as an exercise to the reader, determining which configmaps aren't used is a bit painful at the moment.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+#### Infrastructure view
+
+Cluster → Workloads
+
+Spinnaker uses the term “cluster” in the infrastructure view as an all encompassing term for the metadata, instances, and environments that comprise an application. This term is incongruent with the Kubernetes definition of a cluster. The term “workload” aligns Spinnaker with other tools and accurately refers to the set of Kubernetes application resources.
+
+Load Balancers
+
+The load balancer tab of Spinnaker’s infrastructure view is used to display deployed Service and Ingress objects. While updating this to “Services & Ingresses” would bring Spinnaker into alignment with other tools, it would require further tweaking if the category is expanded to include future object types. Unlike Spinnaker’s overloaded usage of the term “cluster”, the load balancer label is not applied in direct conflict with a pre-existing Kubernetes specific term.
```
dbyron0: I think you're proposing to leave Load Balancers alone ... which is fine with me, but not totally clear.  It's also in a section called `Terminology Changes`, so if it's not a change, it's a little confusing.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
dbyron0: Aaah, I see you've addressed this at least partially below: https://github.com/spinnaker/governance/pull/94/files#diff-2952773823cfaff034e18ebc1ed38686R248.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Config Map / Pod Relationship
+
+Phase 1 will expose Config Map objects. Phase 2 can surface the association with related Pods and detail what data, such as environment variables, is being used by a Pod.
+
+## Secrets
+
+Secrets are a volume type that can be used to store sensitive information. Due to the similarities they share with Config Maps, it should be possible to extend the Phase 1 UI to display them. Some existing tooling in the Kubernetes space elects not to show data field values. Before exposing these in Deck, it will be necessary to examine Spinnaker’s current authorization architecture [(fiat)](https://github.com/spinnaker/fiat) and design a secure means of rendering sensitive information directly in the UI.
```
justinrlee: I think as a phase one, it may make sense to expose information about the secret (name), but not actual contents of the secret.  I think the part that is immediately relevant is knowledge that the secret exists and where it is.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+*The Job metadata panel as seen in the infrastructure UI*
+
+#### Proposal
+
+##### Display Job metadata on the infrastructure view
+
+The Job metadata panel serves as an effective means of surfacing the status and attributes of deployed jobs. However, it cannot be accessed directly. The metadata panel is not visible to users landing on the infrastructure route. Users must follow a link from a pipeline execution or  manipulate the URL by hand to find a given job. If the job was not deployed using Spinnaker or the execution history is unavailable, no link will exist to view the Job UI. Feedback from Spinnaker users, reveals that there is an expectation that the Job UI panel would appear in the infrastructure view. 
+
+The infrastructure view can be updated to display a list of jobs deployed to each cluster. Selecting a job from this list would then display the job metadata panel. This change would increase the visibility of deployed jobs and make it easier for users to monitor job resources across multiple clusters within Spinnaker.
+
```
justinrlee: Question: I'm assuming that this UI will apply to batch resources deployed both by the deploy(manifest) stage and the run job (manifest) stage, right?

Request: If we are going to do this, can we ensure that in the UI, related jobs are associated together?  For example, if you are using the run job manifest stage, it will annotate a hash to the run job name, so it's technically a different run job - can we group these together (either via annotation, or something else?).

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+### Daemon Sets / Stateful Sets
+#### Current Functionality
+
+Daemon Sets / Stateful Sets are currently rendered in Spinnaker’s infrastructure view. They leverage a similar UI to Replica Sets and successfully convey all key object attributes. Given that the Daemon Set / Stateful Set UI has near parity with other tooling, I recommended leaving this functionality unchanged for this project.
+
+### Namespaces
+#### Current Functionality
+
+There is no dedicated rendering of Namespace metadata or associated Resource Quotas in Spinnaker’s UI. Namespace information is only provided as a string identifying where a deployed object resides. While it would be possible to add functionality for exposing associated Resource Quota objects, this feature would be primarily of use to cluster administrators. Refactoring the Spinnaker UI to surface Namespace resource allocation would be of limited use to application teams.
+
```
justinrlee: Is there any discussion of displaying non-service/ingress load balancers?  I'm specifically referring to the various Istio and similar service-mesh type resources - are there any plans to be able to see the resources backing, for example, a VirtualService?

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+- The Kubernetes SIG
+- Release Notes
+
+## Phase 2
+
+The scope of this project’s first phase has been detailed above. There exists a set of additional Kubernetes objects that may be helpful to expose to Spinnaker users. This section details a second phase of possible enhancements.
+
+## Custom CRDs
+
+Spinnaker currently offers [extension points](https://www.spinnaker.io/guides/developer/crd-extensions/) for advanced users seeking to accomplish a deeper integration with their own Custom Resource Definitions. This functionality offers users the ability to override the default model class for a CRD, implement a custom naming strategy, or offer a custom UI. To accomplish this, users must be comfortable working in Spinnaker’s code base and running a layered build. The complexity of this approach may act as a significant obstacle. Approaches such as general resource browser may present a more accessible alternative.
```
ajordens: The plugin effort does aim to make this style of extension easier and not dependent on a layered build — it may not completely solve the complexity problem tho.

FYI @robzienert @christopherthielen @claymccoy

ajordens:
Echo the comments around this being an excellent rfc, thanks for the diligence @louisjimenez!

Looking forward to continued k8s improvements. Will be interesting to see if/when/how some of these patterns roll out to other cloud providers (in particular re-naming of core concepts/terminology).

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
ethanfrogers: See the `Custom CRDs` section. Istio and the like would all fall into this category.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
louisjimenez: Interesting, I have seen other tooling follow this pattern. I believe this would resolve some (most?) of the security concerns but I think it would be helpful to dig in a bit further. I've added this to the [SIG agenda](https://docs.google.com/document/d/1db_yw1uru99Byvin4lQgm7aUZ9xMmvsARtEiiNUrmBo/edit#heading=h.rs4wkuo2ejno) so we can discuss further.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
dbyron0: Yeah, if kubernetes had native GC, I wouldn't be so excited for spinnaker to help :)

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
```diff
@@ -0,0 +1,262 @@
...
+- Render Persistent Volume Claims in the infrastructure view of Spinnaker’s UI
+  - Expose PVC specific metadata covering the access mode, storage class, and a reference to the underlying volume.
+  - Introduce a “Storage” infrastructure tab for rendering cluster level objects such as PVCs.
+  - Link to the underlying Persistent Volume for each PVC.
+- Render Persistent Volumes under the new Storage tab
+  - Expose Persistent Volume specific metadata on the reclamation policy, source disk, and storage capacity.
+  - Link back to the underling PVC.
+- Update the Pod UI to link to the associated PVC
+
+![](storage-mock.png)
```
duftler: Just wanted to call out that there is an alternate scheme used to add additional tabs when "Canary" is enabled for a given application:

Before:
image


After:
image
image

Maybe we can employ some alternate grouping for these newly-rendered resources as well.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
justinrlee: So to my understanding, there are two types of `batch` objects in the Kubernetes API: Jobs and CronJobs

Independent of the above, I think I'm primarily concerned about batch objects where there may be more than one instance of the job running at a given time, and I think there are two high-level use cases for this:
* A pool of worker pods churning through some pool of jobs
* A given job that may be called from different pipeline executions (same job, different parameters; for example, each 'job' creates and monitors a ticket, which is a use case I've seen a lot recently)

Key point here: Job specs are immutable, so the way the RJ stage works is that Spinnaker 'versions' the RJ with an appended string.

I think CronJobs can be left for a later discussion - I think an ideal goal would be something like the following:
* CronJobs create and manage Jobs, similar to the way Deployments create and manage ReplicaSets, so you could have a 'parent' CronJob object that has child Job objects, each of which would have child Pods
* I think this would require building a bit of an abstraction layer on top of the CronJob object in Spinnaker (haven't dug too much into this), so maybe this is a longer-term goal, just something to keep in mind.

For (again, immutable) Jobs created via the "Deploy (Manifest)" stage, I think the behavior is relatively straightforward - group all of the pods for a given job together the same way the standalone un-versioned ReplicaSet object is displayed (with the knowledge that in many instance, there will probably only be one pod running at a time).

For Jobs created by the "Run Job (Manifest)" stage, Spinnaker currently appends an alphanumeric string to the job `name` before deploying it.  If there were a way to track the un-appended job name in the spec (annotation of some sort - `moniker.spinnaker.io/job`?), then we could have all the pods for a given job instance in one level, and then all the Jobs for a given job grouped together at a higher level.

I'm not sure I'm doing a great job of explaining this, but hopefully this makes sense.

justinrlee:
Btw - echo the comments above - this is a fantastic RFC, and I'm looking forward to the work that comes out of it!

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
duftler: Could include copy-to-clipboard links with generated `kubectl` commands to decode secrets for users that have `kubectl` and the required permissions.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
louisjimenez: Thanks, I hadn't seen the Canary specific UI. I'm going to take a look at it.

louisjimenez: @duftler and I chatted about this. The Canary UI relabels the Pipelines link and turns it into a menu. In this case, the infrastructure view will continue functioning as a menu and additional menu items will be added for Storage and Config.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
louisjimenez: Thank you for the detailed feedback!

+1, Displaying the Jobs and Pods that were triggered by a Cron Job would be helpful. This may be achievable by grouping on the ownerReferences metadata available on Jobs triggered by Cron Jobs. 

Yeah, I think Jobs deployed through the Deploy Manifest stage should group pods together in the UI. I plan on having parallel jobs render their associated pods.

I like the idea of grouping the unique Job objects created by the Run Job Stage and leveraging annotations seems like a viable way to do so. I’m not as familiar with the internals of how those annotations are managed so this may end up being a follow up to my initial set of changes.

I’ve gone ahead and updated the RFC to reflect all this.

rfc/kubernetes-ui-enhancements/kubernetes-ui-enhancements.md
louisjimenez: Update on the SIG discussion. The agenda notes are [here](https://docs.google.com/document/d/1db_yw1uru99Byvin4lQgm7aUZ9xMmvsARtEiiNUrmBo/edit#heading=h.rs4wkuo2ejno). The main takeway was that displaying secret metadata alongside the keys (not values) would be of use. I'm leaving this section unchanged just to manage the scope of the project but I think this points to a good path forward for displaying secrets.

louisjimenez:
Looking to merge this but always open to more feedback. @ajordens @robzienert If this looks good, could you mark as approved? :smile:

PR #100, opened 2020-02-04
docs(rfc): Redis removal

robzienert:
I just threw out a timeline that wouldn't shake the world up too much.

### DIFF
--- a/rfc/redis-removal.md
+++ b/rfc/redis-removal.md
@@ -1,0 +1,171 @@
+# Redis Removal
+
+| | |
+|-|-|
+| **Status**     | Proposed |
+| **RFC #**      | [100](https://github.com/spinnaker/governance/pull/100) |
+| **Author(s)**  | Rob Zienert ([`@robzienert`](https://github.com/robzienert)) |
+| **SIG / WG**   | Platform SIG, Operations SIG |
+
+## Overview
+
+To help reduce the technical debt of the project and improve the operational
+overhead of running Spinnaker, Redis needs to be deprecated and removed from
+being a persistence option in favor of using SQL as the baseline storage
+technology. This RFC documents a process and timeline for doing so.
+
+### Goals and Non-Goals
+
+Goals:
+
+*   Provide an inventory of Spinnaker services that still need SQL support added
+*   Define a process that all services can use for migrating datastores without
+    downtime
+*   Provide a timeline for the final switch to SQL, along with criteria to
+    decide if the timeline should be delayed
+
+Non-Goals:
+
+*   Debating if SQL is the best replacement default persistence backend to
+    standardize on
+*   Documenting and solving all technical issues for each service to migrate;
+    those will be handled separately whereas this document concerns itself
+    primarily with the larger processes and timeline
+*   Identification of all new processes for maintaining services on a SQL
+    backend; these will be handled separately
+*   Prescription of what SQL vendor to use: Current functionality has been built
+    using MySQL; support for other vendors will be addressed separately
+
+## Case Against Redis
+
+Redis was introduced into the project as an easy way to get started. It's
+phenomenally flexible, and a single, common data store simplifies the
+operational story of the large operational footprint that Spinnaker has.
+Further, it was useful in that its flexibility meant services could be rapidly
+iterated on, as no schema was needed from the start.
+
+Unfortunately, Redis is difficult to operationalize as a persistent datastore,
+which many of the services treat Redis as. Community members are often surprised
+when Redis restarts to find Spinnaker has lost in-flight work. Operating Redis
+beyond a cache is not a skill that many organizations have. Worse yet, the level
+of effort to support a highly-available, persistent story for Redis is very high
+and littered with edge cases. For example, when tested at Netflix, both Redis
+Cluster and Sentinel regularly encountered incorrect master failover behavior
+that would lead to corrupted clusters or lost data.
+
+Furthermore, developing applications on Redis is typically a foreign concept and
+offers a confusing touch point for anyone looking to get involved in
+contributing.
+
+Having a singular datastore for Spinnaker is still necessary, as it eases the
+burden of onboarding. Now that Spinnaker is no longer a young project, the need
+for fast service iteration has largely dissapated. SQL was chosen as the
+replacement because modern SQL is flexible and well understood.
+
+## Service Inventory
+
+Most of the larger services already have support for SQL.
+
+* Full support: ✔️
+* Partial support: ~
+* No support: ✗
+
+| | |
+|-|-|
+| Clouddriver | ✔️ |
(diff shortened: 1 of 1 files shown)


rfc/redis-removal.md
```diff
@@ -0,0 +1,171 @@
...
+
+_Some words on how services should be written such that they can be migrated
+without downtime._
+
+Being able to migrate existing deployments from Redis to SQL for each particular
+service will be crucial. Each service must provide these features and
+characteristics before they can be said to having "full support":
+
+* A service must support a dual-backend mode, where writes are sent to the SQL
+  store while reads can still hit Redis
```
plumpy: I don't think this is going to work? If you're not writing to both places, then any reads that go to Redis are just going to serve you some random stale data... whatever was in there when you turned off writes.

rfc/redis-removal.md
```diff
@@ -0,0 +1,171 @@
...
+
+Most of the larger services already have support for SQL.
+
+* Full support: ✔️
+* Partial support: ~
+* No support: ✗
+
+| | |
+|-|-|
+| Clouddriver | ✔️ |
```
plumpy: Is it your belief that every cloud provider will work with the MySQL backend? That was not my impression, but I'd be happy to hear it...

rfc/redis-removal.md
```diff
@@ -0,0 +1,171 @@
...
+
+*   Debating if SQL is the best replacement default persistence backend to
+    standardize on
+*   Documenting and solving all technical issues for each service to migrate;
+    those will be handled separately whereas this document concerns itself
+    primarily with the larger processes and timeline
+*   Identification of all new processes for maintaining services on a SQL
+    backend; these will be handled separately
+*   Prescription of what SQL vendor to use: Current functionality has been built
+    using MySQL; support for other vendors will be addressed separately
```
plumpy: Okay, but as long as the messaging makes clear that you have to switch to _MySQL_, not just _some SQL backend_ because that's the truth of the matter and the distinction is likely to be important to some customers...

I've seen people in Slack asking why they couldn't get it working with Oracle, which seems a little naïve, but I do feel like our docs are partially to blame since they refer a lot to "SQL" when I'd expect it to say MySQL if you only support the one, as we do.

rfc/redis-removal.md
```diff
@@ -0,0 +1,171 @@
...
+* A service must provide a clear log signal of when migration is complete
+
+_** Caches and short-lived data do not need to be migrated_
+
+## Timeline
+
+This migration is assuming a slow migration cadence for the rest of the
+Spinnaker services.
+
+### 1.20 (early May)
```
plumpy: Does "complete technical tasks" imply that the service migration process listed above exists for the service? If so, does that process already exists for `clouddriver`, `front50`, `keel`, and `orca`?

rfc/redis-removal.md
```diff
@@ -0,0 +1,171 @@
...
+| Front50 | ✔️ |
+| Gate | ~ |
+| Igor | ✗ |
+| Kayenta | ~ |
+| Keel | ✔️ |
+| Orca | ✔️ |
+| Rosco | ✗ |
+| Swabbie | ✗ |
+
+## Service Migration Process
```
plumpy: Any thoughts about testing processes for this? This feels like potentially the trickiest technical part, and also the part that is hardest to test, and also the part that is most likely to cause catastrophic issues if it goes wrong. So it makes me feel 😱

Maybe there is a tool that can round-trip the data from Redis to MySQL and back, and then verify that the data is completely unchanged. Then we could tell customers "if you can run this round-trip tool on your Redis database, you can trust that we'll migrate it safely"? Just a first idea, maybe not a great one.

rfc/redis-removal.md
robzienert: For clouddriver, orca and front50 we've been sending new writes to the new system, then reading from both the old & new. Seems to have worked fine so far. Then the background worker moves over any data not yet in the new datastore.

rfc/redis-removal.md
robzienert: Ah, you're right. There are a couple things that need to be ironed out, but other cloud providers _can_ go onto SQL. A few people have managed to get k8s on SQL working.

rfc/redis-removal.md
robzienert: Good point, I'll update so it's clearer.

rfc/redis-removal.md
robzienert: Yes, that is what it implies for all of those services - Keel excluded as it's only ever been MySQL. 

In fact, we've just tested the migration process again (SQL to SQL this time). These migration tools and processes are how we first moved away from Redis as well. Since we can't take downtime, we've been pretty deliberate about getting a good migration story in place for the services we've refactored so far.

rfc/redis-removal.md
robzienert: Testing tools would be good, yeah. We've done the other services with unit & integration tests, but for the most part did things yolo after a few practice runs in a lower environment.

The round-trip tool is interesting an interesting idea. It'd definitely give people more confidence into the migration process. I might give thing a try on one of the smaller services to see how it feels. 💯

rfc/redis-removal.md
ezimanyi: By "reads can still hit Redis" do you mean all reads *will* hit Redis (as well as MySQL).  I think you're describing what `DualExecutionRepository` does in Orca where it hits both repositories for all reads and returns the superset (roughly speaking)...but the way it's written makes it seem some reads will only hit redis.

rfc/redis-removal.md
ezimanyi: I know that there were some issues that had to be fixed to get Kubernetes to work, so it wouldn't surprise me if similar issues were seen with other cloud providers.

rfc/redis-removal.md
robzienert: Yes, that's what I meant. I'll make it clearer.

rfc/redis-removal.md
```diff
@@ -0,0 +1,171 @@
...
+* Full support: ✔️
+* Partial support: ~
+* No support: ✗
+
+| | |
+|-|-|
+| Clouddriver | ✔️ |
+| Echo | ✗ |
+| Fiat | ✗ |
+| Front50 | ✔️ |
```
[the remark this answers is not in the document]
robzienert: There is, yes, however it's in the `front50-sql` module so not entirely obvious:

https://github.com/spinnaker/front50/blob/master/front50-sql/src/main/kotlin/com/netflix/spinnaker/front50/migrations/StorageServiceMigrator.kt

robzienert:
Closing, as I don't have a lot of time to be point on moving this initiative forward from a community perspective. I will reopen later when I do have time or there is community interest.

PR #108, opened 2020-02-28
chore(docs): Documenting manual steps involved in SIG creation

robzienert:
Some internal docs for the TOC. Have I missed anything?

### DIFF
--- a/committee-technical-oversight/sig-creation.md
+++ b/committee-technical-oversight/sig-creation.md
@@ -1,0 +1,11 @@
+# SIG Creation
+
+Manual steps involved in creating a SIG.
+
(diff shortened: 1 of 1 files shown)


committee-technical-oversight/sig-creation.md
```diff
@@ -0,0 +1,11 @@
+# SIG Creation
+
+Manual steps involved in creating a SIG.
+
+- Setup Mailing Lists
+  - `sig-{name}@spinnaker.io`
```
plumpy: Mention that `sig-{name}-leads` should be added as member of the group or is that too obvious to point out?

PR #110, opened 2020-03-09
feat(rfc): add website RFC

brian-armory:
This RFC proposes the work the Documentation SIG is taking on to create a revamped Spinnaker.io.

I'll provide high-level updates at the monthly SIG Leads/TOC sync.


rfc/website-revamp.md
```diff
@@ -0,0 +1,115 @@
...
+    * Usage stats
+    * Governance & SIG info. This should be autogenerated from the Governance repo if possible. If not, we will continue to link to the Governance repo.
+* News
+    * Spinnaker newsletter
+    * Featured videos
+* Blog
+
+### The refreshed home page should:
+* Answer the question “Why Spinnaker” at high level and provide easy links to segmented answers (dev, c-level, director, partner etc)
+  * Individual Contributor: What does my life look like if I use Spinnaker?
```
maggieneterval: Is `Individual Contributor` here meant to be `End User`? Perhaps it would be worthwhile to add sections that:

1. More explicitly define the personas you reference in this RFC, and
2. More explicitly define the new entryways and paths that correspond to each persona.

From my reading of this RFC, we are considering the following personas:
- Individual Contributor: I am a developer who contributes code to the core Spinnaker project, or maintains a fork or custom extensions of Spinnaker for my company.
- Operator: I manage a Spinnaker installation for my company.
- End User: I am on an application team and use Spinnaker to manage my continuous delivery workflow.
- Non-technical stakeholder/C-level: I am evaluating Spinnaker as a means to achieve some business objective.
- Evaluator: I am currently evaluating Spinnaker for my company. I could be either an operator, end-user, or higher-level less technical stakeholder. This is almost a dimension of the other personas rather than its own persona, but since the first item in the `Goals` section refers to a persona of `someone brand new to Spinnaker`, might be worth nailing down which "new" personas we are targeting here.

brian-armory: Thanks for the feedback @maggieneterval! We'll work on adding this info to the RFC. Some of it already exists in our notes.

rfc/website-revamp.md
```diff
@@ -0,0 +1,143 @@
...
+* Do nothing
+* Start completely over
+
+## Known Unknowns
+What do people outside of the Docs SIG want to see in a new Spinnaker.io?
+What is the impact of Spinnaker.io on the contributor experience and Spinnaker adoption?
+
+## Security, Privacy & Compliance
+
+n/a
```
ezimanyi: The above section notes that we'll start including usage stats on the site; might be worth commenting on that in this section (even if just to link our to the stats collection RFC).

rfc/website-revamp.md
```diff
@@ -0,0 +1,143 @@
...
+    * Slack join card
+
+* Individual Contributor track: Should answer "What are my next steps?" for potential contributors
+  * Contributor card
+    * leads to Contributor section in the Community area
+  * Spinnaker Universe Calendar card
+
+### Documentation Features
+
+* Versioned documentation that corresponds to Spinnaker versions. This is a small widget in the top nav. See [K8s docs](kubernetes.io/docs/home) for an example.
```
ezimanyi: I think versioned documentation would be great, as we often just add inline to the documentation "you must have Spinnaker 1.x.y to use this feature", but I'd like to understand a bit better how this would be technically implemented.

I think versioned documentation is easiest in cases where the docs are auto-generated by the code, but most of our docs are not.  What would the experience look like for someone writing docs going forward?  Would there be multiple branches in the docs site and contributors would need to update the relevant branches depending on the docs they are writing?  (That is, if you're adding missing documentation for an old feature, you'd need to cherry pick that back, or something similar?)

rfc/website-revamp.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Goals
+
+The ultimate goal is a more modern looking and consistently styled spinnaker.io.
+
+To that end, these are the specific goals we are working toward:
+
+* Provide clear entryways and paths for the different personas spinnaker.io should speak to. For example, a more clear and opinionated navigation for someone brand new to Spinnaker.
+* Provide a clear and compelling answer to anyone asking, “Why Spinnaker for Continuous Delivery?”
+* Add useful documentation features, e.g. versioned documentation and a better in-page table of contents
```
ezimanyi: This is maybe way too specific a request for the RFC, but I'd love to see a search bar on the landing page for spinnaker.io, as  I always need to click on some random link to get the search bar before I can search for something.

rfc/website-revamp.md
```diff
@@ -0,0 +1,143 @@
...
+We need to provide context to attract new companies and ICs as they encounter Spinnaker. This aligns with Travis Tomsu’s efforts to add more context and persuasive oomph to READMEs across the project. The window for Spinnaker inevitability is threatened by a crowded marketplace and will close; this project will boost Spinnaker’s profile and chances for inevitability.
+
+From a technical standpoint, the way GitHub pages manages themes has changed to remote themes, and our implementation is outdated. Relying on the Minimal Mistakes theme-maintainer’s repository results in less theme-related maintenance for the Docs SIG and capture of theme feature improvements.
+
+## Timeline
+* Achieve buy-in from stakeholders by March.
+* Finalize Statement of Work and share with community in March.
+* Finalize SIG-validated high-level design in April.
+* Sitemap
+* Accept contractor bids, including billing rate and time estimate, in April.
```
ezimanyi: Has there been discussion around what this contractor process would look like, and where the funds will come from to pay for it?  I don't know of any precedent for this in the Spinnaker project, so would like to make sure this is fully vetted by the TOC/SC and probably also the CDF before we move forward.

brian-armory: We discussed this in the TOC/SIG leads sync. The conclusion we came to was that @plumpy is going to help @dnilasor ask the CDf for $$$$ once this RFC is approved.

rfc/website-revamp.md
brian-armory: Proprietary documentation w/ versions typically maintain all versions that haven't been deprecated. K8s maintains only the most recent minor version, but the previous versions are still available for people to reference. That seems like a small but useful first step for us. 

I'll add the details on how we would implement this in the RFC.

rfc/website-revamp.md
brian-armory: Thanks! We don't envision any new stats than what's already on spinnaker.io, so linking to the stats collection RFC should be good enough. Will add.

rfc/website-revamp.md
brian-armory: Agreed! Will add. I think we took this for granted and assumed it'd be there.

rfc/website-revamp.md
dnilasor: @ezimanyi @brian-armory I synced with @plumpy on this and am reaching out to Dan Lopez about it today to start the conversation as it may take a while to fully capture his attention. Agree it should be vetted, and if possible, sponsored by the CDF.

rfc/website-revamp.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Design
+
+### Spinnaker Personas
+* Individual Contributor: I am a developer who contributes code to the core Spinnaker project, or maintains a fork or custom extensions of Spinnaker for my company.
+* Operator: I manage a Spinnaker installation for my company.
+* App Developer: I am on an application team and use Spinnaker to manage my continuous delivery workflow as an end-user.
+* Non-technical stakeholder/Executive: I am evaluating Spinnaker as a means to achieve some business objective.
+
+### Top Nav Menu
```
robzienert: There's an effort as part of Spinnaker as a Platform (plugins, etc) that will auto-generate reference docs and samples from code. My strategy here is that we'll be able to generate and export Jekyll content that could be consumed by the spinnaker.io repo.

Would this class of documentation just go under "Docs" under a sub-category of "Reference" or something like that? This is a totally new way of providing docs into the website, so I'm just curious on thoughts. Since the docs will be served from outside locations, it seems to me that it should be somewhat segregated in the navigation so that we don't accidentally stomp on docs written against the main repo.

rfc/website-revamp.md
```diff
@@ -0,0 +1,148 @@
...
+    * Governance & SIG info. This should be autogenerated from the Governance repo if possible. If not, we will continue to link to the Governance repo.
+* News
+    * Spinnaker newsletter
+    * Featured videos
+* Blog
+
+### Entrypoints & Tracks:
+Check out [the mock-up draft](https://docs.google.com/drawings/d/1FCzzvnK0a-_dQ5Nn1T4jz4DrXfr4tMzc9BrObMdEw3k/edit) to visualize the proposed content.
+
+*Note: tracks overlap just as personas overlap for real individuals, based on roles and experience*
```
robzienert: +1 I like this track concept!

rfc/website-revamp.md
brian-armory: It would go in Docs > Reference. We are planning to have that also be the location of the current Halyard command reference, which is also autogenerated from my understanding. Parallel to this website project, @aimeeu is also working on a docs contributor guide where we can note all the auto generated docs.

aimeeu: Yes - working on contributor docs and a SIG Docs release handbook based on my Kubernetes SIG Docs experience. Right now both are low priority, though,  as I work on plugin and operator docs. Thanks for everyone's patience!

robzienert: Great news! Thanks!

dnilasor:
@plumpy good thinking. I have added analytics updates, the 2 that I can think of. I'd like to start using tag manager to track activity. Also seems like a good time to do whatever the CDF needs us to do to consolidate their project web assets. I will ask Dan about his preferred strategy (moving the property or creating/adding to a new one) for more details on the work required.

dnilasor:
Updated this with an updated version of the SOW I recently shared with the CDF budget committee. This provides a clearer work breakdown, and an explanation of what we'd like to get help with, versus what we'll do ourselves. @aimeeu and @brian-armory and approvers, please add anything that I have missed! The more granular we can be now, the better. And since we are going to receive budget one way or another, this is a good opportunity to get your docs and website nice-to-haves.

PR #111, opened 2020-03-13
feat(rfc): add Legacy Artifacts UI Removal RFC

maggieneterval:
Adds RFC to remove the legacy artifacts UI.

### DIFF
--- a/rfc/legacy_artifacts_ui_removal.md
+++ b/rfc/legacy_artifacts_ui_removal.md
@@ -89,20 +89,24 @@ artifacts configured in Produces Artifacts sections of stages.
 
 - [] Document and resolve all bugs discovered while manually testing the new UI.
 
-- [] For release 1.20, remove Deck's `artifacts` and `artifactsRewrite` feature
-flags from both Deck and Halyard. The default UI will be the new UI,
+- [] For release 1.20, remove the `artifacts` and `artifactsRewrite` feature
+flags from Deck. The default UI will be the new UI,
 previously gated by the `artifactsRewrite` flag. Users can revert to the legacy
 UI by enabling a new temporary feature flag, `legacyArtifactsEnabled`. We will
 not expose this flag in Halyard and require it be added manually to
-`settings-local.js`. Communicate this change by updating the artifacts
-documentation and adding a section to the 1.20 release notes.
+`settings-local.js`. Communicate this change by adding the
+`@ValidForSpinnakerVersion` annotation to the `artifacts` and `artifactsRewrite`
+feature flags in Halyard with an appropriate `upperBound` and `tooHighMessage`,
+updating the artifacts documentation, and adding a section to the 1.20 release
+notes.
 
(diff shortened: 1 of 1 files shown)


rfc/legacy_artifacts_ui_removal.md
```diff
@@ -0,0 +1,131 @@
...
+UI_:
+
+- Legacy UI: Enabled by setting Deck's `artifacts` feature flag to `true`.
+Disabled by default. The
+[documentation](https://www.spinnaker.io/reference/artifacts/) currently
+refers to this UI as the "standard UI", but it has also been referred to as the
+"pre-rewrite UI."
+
+- New UI: Enabled by setting Deck's `artifactsRewrite` feature flag to `true`.
+Disabled by default. The
```
dbyron0: Sorry for the annoying question, but if both the legacy and new UIs are controlled by feature flags (and disabled by default), what do you get when both are disabled?

maggieneterval: Not an annoying question at all, David! If both feature flags are disabled (which is currently the default), there is no Expected Artifacts section when configuring a pipeline, and the Find Artifact from Resource stage and Find Artifact from Execution stage are not available.

rfc/legacy_artifacts_ui_removal.md
```diff
@@ -0,0 +1,131 @@
...
+workflow, test that pipelines configured with the legacy UI still work with the
+new UI, and that pipelines configured with the new UI work as expected. This
+includes artifacts configured in the pipeline configuration's Expected Artifacts
+section, artifacts configured as expected artifacts inline in stages, and
+artifacts configured in Produces Artifacts sections of stages.
+
+- [] Document and resolve all bugs discovered while manually testing the new UI.
+
+- [] For release 1.20, remove Deck's `artifacts` and `artifactsRewrite` feature
+flags from both Deck and Halyard. The default UI will be the new UI,
```
plumpy: What happens when you remove them? Will there be an error? It would probably be nice if there was something that said "you were using the `artifacts` feature, but that's deprecated now, the new thing is available instead, or you can use `legacyArtifactsEnabled` to turn the old crappy thing back on (but must migrate before Spinnaker 1.21)!"

I guess maybe that happens automatically via the `tooHighMessage` parameter to `@ValidForSpinnakerVersion`? If so, that's great, maybe mention that will happen under "Community this change by..."?

maggieneterval: Thanks for calling this out, @plumpy! I updated the RFC to clarify that we will annotate the existing artifacts feature flags with `@ValidForSpinnakerVersion` with an appropriate `tooHighMessage` for release 1.20, and only remove them after release 1.23.

maggieneterval:
@robzienert Does this look okay to you to merge? @ethanfrogers is on paternity leave so I think yours is the last remaining vote. 😄

PR #115, opened 2020-03-30
Design document for configuring logo in default spinnaker login page.

vvtelu:
https://github.com/spinnaker/gate/pull/1004


plumpy:
I think the PR is fine for this. It's a pretty small PR and this doesn't really add much, so I'm going to close it. Thanks.

PR #129, opened 2020-04-23
feat(rfc): update Kubernetes V1 and Legacy Artifacts Removal RFCs

maggieneterval:
* feat(rfc): update Kubernetes V1 EOL RFC 

* feat(rfc): update Legacy Artifacts Removal RFC 

* feat(rfc): mark Kubernetes V1 EOL RFC as implemented


maggieneterval:
Just added a third commit to mark the Kubernetes V1 EOL RFC as implemented.

PR #141, opened 2020-05-13
feat(rfc): close out RFCS for EOL of Kubernetes V1 provider and legac…

maggieneterval:
…y artifacts UI

I've assigned issues to myself to track the long tail of cleanup work related to these RFCs, but barring any unforeseen issues, we can now mark them implemented. I've also added a Halyard-specific table to the timeline section of each RFC to clarify the timing of each Halyard milestone by Halyard release rather than Spinnaker release.

### DIFF
--- a/rfc/eol_kubernetes_v1.md
+++ b/rfc/eol_kubernetes_v1.md
@@ -42,7 +42,7 @@ The Spinnaker Kubernetes SIG and Slack channels operate best with a common found
 
 ## Timeline
 
-Each set of milestones should be complete by the time the corresponding Spinnaker version is released.
+Each set of milestones should be complete by the time the corresponding Spinnaker or Halyard version is released.
 
 | Spinnaker Version | Release Window Opens | Milestones |
 |-------------------|----------------------|------------|
@@ -50,6 +50,12 @@ Each set of milestones should be complete by the time the corresponding Spinnake
 | 1.19              | 2020-03-02           | Publish a post to the Spinnaker Community Blog explaining why we are removing V1 and how to migrate to V2. Add a feature flag that must be enabled before deploying to V1 accounts. Add appropriate warnings to Halyard, Clouddriver, and Deck. Highlight in 1.19 curated changelog that V1 will be removed in 1.21.
 | 1.20              | 2020-04-27           | Complete audit of community docs and blog posts for V1-related content to update. Highlight in 1.20 curated changelog that V1 will be removed in 1.21.
 | 1.21              | 2020-06-22           | Remove V1-only code from all Spinnaker microservices.
+
+| Halyard Version | Supported Spinnaker Versions | Milestones |
+|-----------------|------------------------------|------------|
+| 1.32            | 1.17, 1.18, 1.19             | Add warning for any configured legacy Kubernetes accounts. Default new and unspecified accounts to standard (V2) provider.
+| 1.36            | 1.18, 1.19. 1.20             | Remove legacy Kubernetes distributed installation path. Spinnaker itself must be deployed using a standard (V2) account.
+| TBD             | 1.21, 1.22, 1.23             | Stop exposing `--provider-version` command, as no supported release will include support for the legacy (V1) provider.
 
 ## Design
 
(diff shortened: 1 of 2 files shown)


rfc/eol_kubernetes_v1.md
```diff
@@ -51,6 +51,12 @@ Each set of milestones should be complete by the time the corresponding Spinnake
 | 1.20              | 2020-04-27           | Complete audit of community docs and blog posts for V1-related content to update. Highlight in 1.20 curated changelog that V1 will be removed in 1.21.
 | 1.21              | 2020-06-22           | Remove V1-only code from all Spinnaker microservices.
 
+| Halyard Version | Supported Spinnaker Versions | Milestones |
+|-----------------|------------------------------|------------|
+| 1.32            | 1.17, 1.18, 1.19             | Add warning for any configured legacy Kubernetes accounts. Default new and unspecified accounts to standard (V2) provider.
```
plumpy, line 56: Do we actually have a policy that versions of Halyard only support n old versions? If you try to use Halyard 1.32 to deploy 1.16, will it give you a warning? I don't think Halyard even knows what version of Spinnaker is the latest at the time it's published, does it?

maggieneterval: The (lack of?) relationship between Spinnaker and Halyard releases has always confused me...I think we only have a policy that at any given time, the OSS community considers the most recent three minor Spinnaker versions supported, and only the single most recent Halyard minor version supported (since I think we only ever do cherry-picks back to the most recent Halyard minor version?). So from that it seems reasonable that a given Halyard minor version should only need to support the three most recent Spinnaker minor versions at the time it was released? But you're totally right, there's no warning if you, for instance, try to use Halyard 1.32 to deploy 1.16, because there's no actual Halyard-knowledge of which versions are supported. We can set a _minimum_ Halyard version for a given Spinnaker version, which doesn't really help us with deprecation cases like this since what we actually want is to set a minimum Spinnaker version for a given Halyard version.

Do we think the policy should actually be "you can use any Halyard version to deploy any Spinnaker version, even versions that were unsupported at the time of the Halyard release"? If yes, then we can just never remove flags that are no longer used in any supported Spinnaker release. The primary downside to this is that we'll be burdening new/prospective users, who consistently report feature flag overwhelm, in exchange for making it easier for folks to upgrade to the latest version of Halyard but stay at an unsupported Spinnaker version, which I'm not sure there's a great story for given that I'm not sure there are any Halyard features in the pipeline even usable with unsupported Spinnaker versions that would make users want/need to use the latest Halyard?

plumpy: > The (lack of?) relationship between Spinnaker and Halyard releases has always confused me...

You're not alone :smile: 

> Do we think the policy should actually be "you can use any Halyard version to deploy any Spinnaker version, even versions that were unsupported at the time of the Halyard release"?

Well, no, I don't think that's what the policy _should_ be. But I don't think we've ever had any policy about it at all before. So while I think what you're describing ("a given version of Halyard only supports Spinnaker releases that were supported at the time of its release") is a totally reasonable one, I do think it's new.

On the one hand, if we were going to establish that policy it would probably be nice to add enforcement for it in Halyard. But maybe it's not worth the effort to make that happen.

maggieneterval: > On the one hand, if we were going to establish that policy it would probably be nice to add enforcement for it in Halyard. But maybe it's not worth the effort to make that happen.

If any additional effort were to be made to improve Halyard, I'd vote for documenting the creation of a release changelog first, since right now there's not a great way (apart from ad-hoc #halyard channel messages) for us to communicate _any_ changes to users. (Not sure we even want to prioritize that above other release process improvements, given Halyard's eventual deprecation. 🤷)

I guess this all stems from the root issue of Halyard bundling management for both 1) configuration for services, which feels like it should be versioned with the services and 2) deploying those services, which should not need to be coupled to versioning of those services, into one big monster that took the decoupled versioning strategy without fully considering how we'd make breaking configuration changes. But I'm preaching to the choir. 😇

PR #154, opened 2020-07-08
A proposal for a consolidated build system

plumpy:
The thrust of it is a proposal for a monorepo for all our Java projects.

It also has a proposal for a new version number scheme (2020.4 instead of 1.23) and a change to the release cadence. Technically those are unconnected to the new build system, but it seemed like as good of a time as any to debate such a change.


cfieber:
I can dig it

I think tidying up and/or adding good CODEOWNERS files in all the repos would be pretty helpful towards maintaining some semblance of sanity when all the PRs are against a single repo

Any specific reason not to roll the spinnaker-gradle-project into this since it would be nice to be able to update that and apply it across all the projects as well (and I suspect we can reference it compositely or as buildSrc instead of requiring a plugin publishing step)

cfieber:
also nice and probably possible would be getting rid of the gradle wrapper checked into each sub project - I suspect if there was a single top level gradle wrapper it would Just Work™ and if necessary we could add gradlew symlinks one directory down so that script still appears to exist in each sub module

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,698 @@
...
+1.22.0 in one of the old version schemes.
+
+The release branches will be cut on the second Wednesday of the quarter, and the
+release will be targeted for the following Wednesday. (These dates were chosen
+to avoid major US holidays.) The time between releases will now be either 13 or
+14 weeks, depending on the calendar. The next releases will be:
+
+- 2020.4: branches cut 2020-10-14, release on 2020-10-21
+- 2021.1: branches cut 2021-01-13, release on 2021-01-20
+- 2020.2: branches cut 2021-04-14, release on 2021-04-21
```
ncknt, line 97: `2021.2` right?

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,698 @@
...
+repository.
+
+### Java Monorepo
+
+All the Java repositories will be merged into a single monorepo called
+`spinjava` using [`monorepo-tools`](https://github.com/shopsys/monorepo-tools).
+A test version of the repository described here
+[is available at `plumpy/spinjava`](https://github.com/plumpy/spinjava). The
+script used to generate it
+[is included as an appendix](#appendix-a-script-to-generate-the-monorepo).
```
robzienert, line 142: Hey this is neat, I like!

robzienert:
> also nice and probably possible would be getting rid of the gradle wrapper checked into each sub project

@cfieber I'm not sure I'm into this. How would you see development flows working? Would developers then use the monorepo as their entrypoint, rather than individual service repos?

cfieber:
> How would you see development flows working?

I think by its nature you'd have to have the whole monorepo checked out. You could potentially have several of these (at the cost of disk space and maybe IDE confusion) if you have a bunch of separate streams of development ongoing and don't want to keep popping between git branches

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,698 @@
...
+projects that want to depend on a recent, unreleased version of the Java
+microservices. For example, `halyard` will use these artifacts, as will users
+developing plugins for the next version of Spinnaker.
+
+Artifacts will be identified by the branch, commit timestamp (in UTC), and short
+SHA (seven digits). They'll look like this: `2020.4-20200601.193201-198aa21`.
+
+Commit timestamps aren't guaranteed to increase monotonically, so technically
+Maven won't always sort these versions correctly. If, for example, you're
+developing a plugin and add a dependency on the latest version of echo, you may
```
ncknt, line 201: For plugins, would we just `require` a Spinnaker version now instead of individual service versions? That would certainly be easier for plugin authors and users.

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,698 @@
...
+#### Publishing a Release
+
+To publish a release of Spinnaker, a release manager will create a GitHub
+release in the `release-operations` repository. The tag used for the release is
+the testing release that will be published.
+
+All published artifacts from the testing release will be updated with the new
+version:
+
+- BOM (for Halyard deployments): The existing BOM will be copied, with the
```
ncknt, line 256: If all services use the same version, we don't need a BOM anymore. `spinnaker-monitoring` and `deck` being the exceptions. Is the plan to eventually align these versions - perhaps just tagging the produced images? Managing BOMs is not fun in some environments.

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,698 @@
...
+This build system will be implemented between the 1.22 release in mid-August and
+the following release (which will no longer be called 1.23) in mid-October.
+
+## Design
+
+### New Versioning Scheme and Release Schedule
+
+Spinnaker releases will switch to date-based versions, so instead of 1.23, we
+will call it 2020.4. The release schedule will change from every 8 weeks to
+every 3 months.
```
ncknt, line 75: I like the stability but we'd probably need to relax the definition of what can be backported to the latest release. I don't think it'd be acceptable to wait a quarter to get a fix that is not technically a regression.

plumpy:
> Any specific reason not to roll the spinnaker-gradle-project into this since it would be nice to be able to update that and apply it across all the projects as well (and I suspect we can reference it compositely or as buildSrc instead of requiring a plugin publishing step)

I guess there were two reasons I didn't include it:
1. I assumed you couldn't include a plugin as a composite build, but I admit I didn't even try
2. it's also used by `deck` and `spinnaker-monitoring` for publishing the Debian packages. So it felt like maybe it shouldn't be tied so closely to the Java code. If we got rid of the Debian packages (still working on a plan for that) then this concern goes away...

But anyway, I'm definitely not totally opposed to the idea...

> I'm not sure I'm into this. How would you see development flows working? Would developers then use the monorepo as their entrypoint, rather than individual service repos?

when you say "individual service repos", you mean "individual service projects", right? Certainly the repositories themselves would be archived or deleted or whatever if we moved them into a monorepo.

But yeah, this was my only real concern about moving the gradle wrapper up a level... it means you either have to do this:
```
echo$ ../gradlew build
```
or this:
```
spinjava$ ./gradlew -p echo build
```

And opening the `echo` directory in IntelliJ doesn't really work either. (It looks for the `gradle/wrapper/gradle.properties` file.)

But having 13 copies of the exact same four gradle wrapper files does indeed seem stupid. So, yeah, maybe symlinks?

I also think there are probably a lot of really good restructurings we could do once we moved into the monorepo. Maybe we _do_ have a top-level `build.gradle` file eventually? I'm kinda proposing the simplest thing first and then we can iterate from there...

rfc/build-system/build-system.md
plumpy: As of this latest release, we actually do tag the containers! `gcr.io/spinnaker-marketplace/deck:spinnaker-1.21.0`, etc. Because of the aforementioned version conflicts, we do have to prepend `spinnaker-` to the tags, since `deck:1.21.0` is something else entirely...

Anyway, that's what I meant (perhaps unclearly) two lines below:

> - Docker: the existing containers for each microservice will be tagged with the new version numbers

It's true that we don't really need the BOMs for containers now (but still do for Debian packages), but I don't really want to bother fixing Halyard to not use them, so they'll probably remain a thing until both Halyard and the Debian packages are fully deprecated.

rfc/build-system/build-system.md
plumpy: Yep, that's one of the main goals of this (after discussions I had with @link108 and @claymccoy): every component of a Spinnaker release (including deck) has the same version number. Your plugin targets Spinnaker 2021.2 instead of some other obscure version number.

rfc/build-system/build-system.md
plumpy: Whoops, thanks! Fixed!

rfc/build-system/build-system.md
plumpy: I updated the bullet to have more context around this, thanks:

> - BOM (for Halyard deployments): The existing BOM will be copied, with the
  included `version` field updated to reflect the new version number.
  (Deployments that aren't using Halyard will just fetch the containers
  directly, but the BOM will continue to be written until Halyard is no longer
  a supported tool.)

rfc/build-system/build-system.md
ezimanyi: Agreed, I think we'd probably need to re-think this in the context of less frequent releases.  I think we should still have a reasonably restrictive policy (not allowing features or overly invasive bugs) but probably would need to allow simple, non-invasive bug fixes even if they do not fix regressions.

maggieneterval: +1. If we only change "must fix a regression" to "must fix a bug," the remaining criteria around exhaustive documentation and test coverage should hopefully be enough to ensure backports are sufficiently simple and non-invasive.

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,711 @@
...
+
+Anecdotally, we believe that some users expect that there won't be any breaking
+changes between 1.20 and 1.21, perhaps confusing Spinnaker versions with
+[SemVer](https://semver.org/). In fact, we regularly introduce breaking changes
+in these releases.
```
ethanfrogers, line 80: Have we ever explicitly stated that release versions aren't SemVer compliant (outside of this doc and calls)? What's the preventing us from following SemVer releases? If it's regularly introducing breaking changes, might there be a better way to solve that problem instead of increasing release cadence? Better testing? Incrementing the major version when appropriate?

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,711 @@
...
+
+- 2020.4: branches cut 2020-10-14, release on 2020-10-21
+- 2021.1: branches cut 2021-01-13, release on 2021-01-20
+- 2021.2: branches cut 2021-04-14, release on 2021-04-21
+- 2021.3: branches cut 2021-07-14, release on 2021-07-21
+
+Slowing the release cadence reduces the release manager overhead somewhat,
+although release manager shifts are now longer and may have to be shared by
+multiple people. It also helps many of our users who find upgrading to be a lot
+of work by reducing how frequently they have to upgrade. It does mean it takes
```
ethanfrogers, line 103: Do we have any data about why users feel like upgrading is a lot of work?

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,711 @@
...
+
+- don't change the current Spinnaker version scheme (this decision really has
+  nothing to do with the rest of the RFC; the build system described here will
+  work with the existing Spinnaker versioning scheme too).
+- Use versions of YYYY.MM, where MM is the month. This makes it even more clear
+  when the release is from, but it looks a little weird with a patch release
+  (2020.07.3). It's also confusing since the first and last components of the
+  version increment by one, whereas the middle component changes by some other
+  interval.
+- Release more often. Release less often. Release every day?
```
ethanfrogers, line 116: With longer release times, it seems like we have an opportunity for features to bake longer, which is nice. The downside is that it takes longer to get feedback from the community. Will there be any kind of `nightly` people can use to take things for a spin (heh)?

rfc/build-system/build-system.md
ethanfrogers: It looks like Kubernetes has a pretty [well documented](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#kubernetes-release-versioning) approach to how they version multiple components using SemVer.

rfc/build-system/build-system.md
```diff
@@ -0,0 +1,711 @@
...
+dependency on `clouddriver`, but building `front50` doesn't require building
+_all_ of `clouddriver`. You just need to build the `clouddriver-google-common`
+module.
+
+Caching works as you'd expect: if you build `orca` followed by `fiat`, you will
+only have to compile each `kork` module once.
+
+This monorepo requires you to change all affected Java code in a single PR; no
+longer will you commit code to `kork` only to find out later it broke
+compilation of a handful of microservices.
```
ethanfrogers, line 166: I'm a fan of this. No more `kork` autobumps!

rfc/build-system/build-system.md
danielpeach: I agree that the Spinnaker version number has never actually signaled semver changes, but I don't think that's a good reason to switch to a different versioning scheme. Users want to know about breaking changes. Can we leave a TODO here and decide, at a later date and in a separate RFC, the scope of Spinnaker's public API and the definition of a breaking change w.r.t semver?

plumpy: > Have we ever explicitly stated that release versions aren't SemVer compliant

What if you flip that around: have we ever explicitly stated that release versions _are_ SemVer compliant? I doubt it. I obviously wasn't there at the time, but I don't think these version numbers were ever intended to be used that way. I think it was more like a "Microsoft Word 5.2"-style version number.

Like Dan points out, I think we'd have to have a documented API before we could even _try_ to implement SemVer. Right now I don't even know what it means. If we remove a configuration option, for example, is that considered a breaking change? If we remove Oracle support that's used by almost no one, do we have to bump the major version?

Honestly, it just doesn't seem like Spinnaker was carefully designed enough to make promises of backwards compatibility. It's a lot of very difficult work to make a promise like that. There are dozens of obscure features implemented with little to no documentation or tests, along with zero metrics to know who is using them, and to hope that we can move the product forward without breaking or removing some of them just doesn't feel realistic to me.

Or maybe we just say "everything right now is beta and we'll mark specific hand-selected things as within the scope of SemVer going forward", but how much value would that actually provide?

> I agree that the Spinnaker version number has never actually signaled semver changes, but I don't think that's a good reason to switch to a different versioning scheme.

I also listed a few other reasons :grin: But yeah, as I mentioned, we don't _have_ to do this now... I'm still inclined to push for it, but not especially vigorously.

danielpeach: > it just doesn't seem like Spinnaker was carefully designed enough to make promises of backwards compatibility. 

I don't think this can be entirely true 😬, though I understand the sentiment. You could start with a small public API scope and expand over time as test coverage improves and users tell us what they care about. I'd love if we could say that pipeline configs are executable across minor releases (with a strong understanding that it's not going to be perfect, a culture of adding tests when you inevitably break minor versions, and a system for marking stage types and cloud providers as alpha/beta/not-covered-by-semver).

rfc/build-system/build-system.md
ajordens: I don't even know what the upgrade process looks like for a typical user looking for zero downtime.

I know at Netflix we try to be aware of the order in which we're upgrading services and even our best attempts are occasionally foiled.

jonsie:
I like the general idea. 

My main concern is regarding library versions.

`kork`, `keiko`, and `-api` versions are effectively Spinnaker libraries.  They are frequently iterated on and I would not expect a full Spinnaker OSS release to be part of the process for releasing versions of these libraries.  There are downstream consumers of these libraries (our `-nflx` extensions, and I believe other folks in the community have built on top of these libraries).  To me, one high level date-based Spinnaker product version makes sense but I still see the need for specific library versions.  Does this make sense?  I don't know exactly how this new system would address this concern.

rfc/build-system/build-system.md
ethanfrogers: > What if you flip that around: have we ever explicitly stated that release versions are SemVer compliant? 

Fair point. I think the answer is no. I _was_ surprised to hear it said out loud the first time as were a few other at Armory and from the community, though. I guess the assumption I made was that "we follow SemVer, just poorly". 


> It's a lot of very difficult work to make a promise like that.

Totally. I can't help but wonder if it's valuable work, though. If anything, understanding all of the points standing in our way - along with the points you mentioned - would give us a clear understanding of how much of a lift it is and whether it's even possible. I do agree with @danielpeach that we should define _some_ degree of backwards compatibility, however big or small it ends up being. Obviously that's outside the scope of this PR, and probably an exercise for the TOC. I'm just exploring the problem a bit. 

I'm happy to provide input/feedback on what that might look like. We (Armory) have a good bit of anecdotal data about how breaking changes affect users and the work they/we have to do to workaround them.

rfc/build-system/build-system.md
ethanfrogers: I'd say that most, if not all, OSS users just run `hal config version edit --version {target-version}` and deploy. Kubernetes does a decent enough job of rotating things without downtime. The majority of the time, an unknown, breaking change doesn't present until things have been successfully deployed.

ajordens: Gotcha. 

Surprise that doesn’t break often (or maybe it does). We still occasionally get tripped up by coupled changes across services. Changes that in isolation don’t look like traditional breaking changes.

rfc/build-system/build-system.md
ajordens: I think it’s worth sharing any data (anecdotal or not). We should create space for an initial discussion on backwards compatibility expectations in the TOC.

plumpy:
FYI to anyone following this, I added a "New Artifact Repositories" section discussing moving from Bintray+Google Container Registry to Google Artifact Registry for storing all our containers, jars, and debian packages.

plumpy:
@jonsie 

> They are frequently iterated on and I would not expect a full Spinnaker OSS release to be part of the process for releasing versions of these libraries.

Well, there would be artifacts published twice-weekly (or some other schedule, I just arbitrarily picked that) as described in the "Maven Artifact Publishing" section. Is that good enough for you? We can also make it so that the action that builds the artifacts could be triggered at any time, so if you wanted a `kork` release you would just run that action and have one in 20 minutes or so.

Would that suffice? I don't really think anyone besides you runs versions of Spinnaker outside the releases, so I don't really think anyone else cares about artifacts aside from those associated with a release (except for doing development targeting the next release).

I know we talked about having a separate version for each of the API modules (i.e. so `orca-api` and `clouddriver-api` would have their own versions) but really I think that just makes it a lot harder for the Spinnaker community, since people just want to write a plugin targeting "Spinnaker 2020.4", not "orca-api 1.8 & clouddriver-api 9.3".

plumpy:
I'd love to discuss all of this more at the [platform SIG tomorrow](https://docs.google.com/document/d/1zGIgEwO9lZfYaI0N6necuH-x0SjmL3hmbbeMWZgqnEw/edit) at 11am US Pacific, 2pm US Eastern! Please attend if you can!

rfc/build-system/build-system.md
cfieber: I feel like for each service's plugin api we could have a legitimate semver story.

Once we monorepo everything I feel it will be harder for plugin developers to express meaningful version constraint requirements on their plugins other than `service>=the release version where I first wrote this` and plugins breaking on version updates will be a riskier and completely untestable by the core project proposition

cfieber:
> since people just want to write a plugin targeting "Spinnaker 2020.4", not "orca-api 1.8 & clouddriver-api 9.3".

I think there are two facets to this. With the spinnaker-wide version it's easier for the plugin author to figure out what version the plugin is targeting when you write it, but it's harder for the operator to tell whether those plugins still work against a newer version of Spinnaker.

I'm strongly in favor of inflicting more difficulty on the plugin author to ease the burden of a spinnaker operator doing an upgrade.

rfc/build-system/build-system.md
alchen99: It's not the upgrade itself that is a lot of work. It's reading or trying to read all the release notes and commits to prepare for upgrading and then the verification that the upgrade didn't regress that is a lot of work.

nicolasff:
Interesting PR, thanks @plumpy!

The monorepo approach targets the specific challenges associated with the current structure and seems to offer a solid alternative that would make builds and dependency management much simpler; this alone is a great improvement.
Without meaning to derail the conversation or go too far off-topic (and speaking for myself and not my employer), I’d like to also offer a slightly different perspective as an engineer on a team that maintains a fork of Spinnaker with a number of custom integrations. Working with the open-source repos of Spinnaker is a breeze compared to the amount of effort involved for any team trying to maintain a sizable fork that tracks multiple release branches in addition to daily changes on `master`.

This is our current approach: we make small changes directly in our clones of the Spinnaker repos (for example back porting a bugfix or adding a few modules to our copy of Kork), while larger features such as custom cloud providers, the stages to drive them, and the atomic operations that go with those are all implemented in dedicated “wrapping” repos. A repo like `orca-wrapper` would depend on packages built from `orca`, and configures Spring Boot to start the Orca service by loading all the built-in Orca components as well as the custom ones written in `orca-wrapper`. Building this repo produces a deployable image of Orca with the added features. Building a significant number of custom stages has meant adopting this structure for Orca, Cloudriver, Deck, Gate, and more.

Writing and maintaining code in these dedicated repos that depend on (potentially modified) clones of the open-source repos, while tracking upstream changes on `master` and bug fixes on the release branches involves a unique set of challenges and requires a significant amount of engineering effort. I realize that this proposal is not about making forks easier to maintain, but it seems to me like it would also have this added benefit. I would wager that making forks simpler to work with could also encourage their maintainers to upstream the features they added; it would certainly be easier in our case.

I’m excited about this proposal and look forward to these changes.

plumpy:
Since Armory has concerns about this monorepo approach, and I'm not particularly interested in forcing the issue, I'm just going to close this RFC.

PR #159, opened 2020-07-30
feat(rfc): externalized account management

ncknt:
Consistent account management across services and providers in Spinnaker.

### DIFF
--- a/rfc/account-management.md
+++ b/rfc/account-management.md
@@ -3,7 +3,7 @@
 | | |
 |-|-|
 | **Status**    | _**Proposed**, Accepted, Implemented, Obsolete_       |
-| **RFC #**      | _Update after PR has been made_ |
+| **RFC #**      | https://github.com/spinnaker/governance/pull/159 |
 | **Author(s)**  | Nicolas Cohen (@ncknt), Nima Kaviani (@nimakaviani) |
 | **SIG / WG**   | Platform |
 
--- a/rfc/account-management.md
+++ b/rfc/account-management.md
@@ -1,0 +1,135 @@
+# Spinnaker Account Management
+
+| | |
+|-|-|
+| **Status**    | _**Proposed**, Accepted, Implemented, Obsolete_       |
+| **RFC #**      | _Update after PR has been made_ |
+| **Author(s)**  | Nicolas Cohen (@ncknt), Nima Kaviani (@nimakaviani) |
+| **SIG / WG**   | Platform |
+
+## Overview
+
+Spinnaker reads account definitions (cloud providers, CI, metric stores, etc.) from Spring properties. This works well in a world with a handful of accounts that rarely change but causes operational pain when provisioning accounts dynamically or with account information stored externally.
+
+This document proposes to add a thin account management layer in each Spinnaker service to support dynamic account management from non Spring data sources.
+
+
+### Goals and Non-Goals
+
+Goals:
+- Provide an interface that each Spinnaker service can optionally implement
+- Support Spring properties (as today)
+- Adding account dynamically without having to restart the service
+- Add stable interfaces that can pull accounts from a custom source
+- Prepare Clouddriver for multiple credential repositories
+
+Non Goals:
+- Storing account information in Spinnaker
+- Migrate all services and all providers
+
+
+## Motivation and Rationale
+
+Spinnaker reads account definitions (cloud providers, CI, metric stores, etc.) from Spring properties. This forces Spinnaker operators to restart services when an account is added, modified, or removed. The community is currently using [Spring Cloud Config](https://spinnaker.io/setup/configuration/) but it is complicated to set up and has significant performance issues as the number of accounts grow.
+
+We're proposing to define interfaces to let each Spinnaker service support a variety of sources of account definitions as well as dynamically reload them.  We'll work in each service to add support - one provider at a time.
+
+With this proposal implemented, operators of Spinnaker could easily:
+- reload accounts dynamically across all types of accounts
+- add a new account source from a specific provider
+- extend the behavior of the `AccountRepository` across all types of accounts.
+- add a new account source and get the account information from its source of truth.
+
+We have evidence from our customers that some form of dynamic account management is needed.
+
+## Timeline
+
+Spinnaker 1.23 (1.22?):
+- Add interfaces to kork 
+- Start splitting AccountCredentialsRepository by provider (we'd now be able to use multiple accounts with the same name in different providers)
+- AWS, Kubernetes support in Clouddriver.
+
+Future releases:
+- More services/providers as the community sees the need. We expect the change to be non invasive.
+
+We have other providers in mind and are ready to contribute for these as well.
+
+
+## Design
+
+Here's a simple implementation of the interfaces (interfaces should be in kork): https://github.com/ncknt/clouddriver/compare/master...ncknt:feat/kork-accounts?expand=1
+
+An `Account` contains the information necessary to access an account. It can be parsed and enriched into a `Credentials`. Spinnaker already has a lot of these: e.g. `com.netflix.spinnaker.clouddriver.security.AccountCredentials`, `com.netflix.spinnaker.clouddriver.artifacts.config.ArtifactCredentials`, `com.netflix.spinnaker.igor.service.BuildService`, etc.
+
+An `AccountParser` translates an `Account` into an instance of `Credentials`. A `CredentialsRepository` contains one type of `Credentials`. Most of the time, we only need access to a single type of credentials (e.g. in clouddriver's providers). When access to different types of credentials is needed (e.g. in clouddriver's artifact), we can use a `CompositeCredentialsRepository`.
+
+Each account repository is wired with an `AccountSource`. The source is queried once when the repository is created but can also be queried again if it is a `ReloadableAccountRepository` which the default implementation is.
+
+If `accounts.dynamic.enabled`, then every `accounts.dynamic.[type of account].reloadFrequencyMs` the source will be queried, parsed, and merged into the repository.
+
+A `CredentialsLifecycleHandler` handles enabling/disabling credentials (e.g. scheduling agents for a new provider account).
+
+Out of the box, nothing should change for users. However, we could now write a plugin that for instance get Kubernetes accounts from a remote endpoint, or polls Jenkins master definitions from a database.
+
+### Clouddriver Providers
+Clouddriver currently uses one repository to store all accounts. Part of the proposal pushes each provider to use a `CredentialsRepository` defined above. It also uses the `AccountCredentialsProvider` abstraction.
+
+For each provider, the migration looks like:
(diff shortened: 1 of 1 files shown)


rfc/account-management.md
```diff
@@ -0,0 +1,135 @@
...
+- More services/providers as the community sees the need. We expect the change to be non invasive.
+
+We have other providers in mind and are ready to contribute for these as well.
+
+
+## Design
+
+Here's a simple implementation of the interfaces (interfaces should be in kork): [https://github.com/ncknt/clouddriver/compare/master...ncknt:feat/kork-accounts?expand=1](https://github.com/ncknt/clouddriver/compare/master...ncknt:feat/kork-accounts?expand=1)
+
+An `Account` contains the information necessary to access an account. It can be parsed and enriched into a `Credentials`. Spinnaker already has a lot of these: e.g. `com.netflix.spinnaker.clouddriver.security.AccountCredentials`, `com.netflix.spinnaker.clouddriver.artifacts.config.ArtifactCredentials`, `com.netflix.spinnaker.igor.service.BuildService`, etc.
```
robzienert, line 62: I'm interested to hear more about why we need a distinction between `Account` and `Credentials`. Do we have an opportunity to simplify things?

rfc/account-management.md
```diff
@@ -0,0 +1,135 @@
...
+### Clouddriver Providers
+Clouddriver currently uses one repository to store all accounts. Part of the proposal pushes each provider to use a `CredentialsRepository` defined above. It also uses the `AccountCredentialsProvider` abstraction.
+
+For each provider, the migration looks like:
+- Move from `AccountCredentialsProvider` to `CredentialsProvider<T extends AccountCredentials>`
+- `CredentialsProvider` has the same methods except it is using generics
+- Create the default `CredentialsRepository` bean
+- Create the default `CredentialsProvider` bean
+
+`clouddriver-elasticsearch` uses credentials across credentials and can be changed to use a `CompositeCredentialsRepository` that federates `CredentialsRepository` with non migrated provider accounts.
```
robzienert, line 83: > uses credentials across credentials

I don't understand this part.

rfc/account-management.md
ncknt: `Account` is supposed to be a POJO that can be read from simple properties. `Credentials` is what services use and may also have injected beans (e.g. `KubernetesNamedAccountCredentials` use a factory. `NetflixAmazonCredentials` uses a `AWSCredentialsProvider`). By separating account source (list of props) from credentials, it becomes easy to create new sources without having to worry about what else is needed.

We could do without `Account` but then each time you want to add a new source, you have to deal with the wiring. Alternatively, we could move the beans outside of the `Credentials` object but that would be a bigger change.

rfc/account-management.md
ncknt: Sorry there's a typo! What I meant is that the elasticsearch module seems to need to access credentials across **providers** - or at least it seemed to but I've never used it so it may very well be wrong. I took it as an exercise: what if we need to access multiple types of credentials? Hence the `CompositeCredentialsRepository`.

rfc/account-management.md
ajordens: I think there's a lot of clumsiness in the current implementation and split between Account/Credentials. It's made deserialization tricky (number of places that need to be changed to introduce a new attribute) and has been a sore spot for some time now.

I don't think it's a terribly clean distinction right now, feels like the wiring could be handled but I've not looked through _every_ implementation.

ncknt: It's true that there's a lot of repetition in some of the providers. Unless we do a major redesign and move all managed beans out of these credentials objects, we will always need to enrich these beans. The same class could optionally represent `Account` and `Credentials` and keeping the same abstraction, `AccountLifecyleHandler` could be in charge of autowiring beans. We could even provide a default handler that does the autowiring.

e.g.

```
class NetflixAmazonCredentials extends AmazonCredentials implements Account {
	...
}

class AwsLifecycleHandler extends AutowiringLifecycleHandler<NetflixAmazonCredentials> {
	
	accountAdded(NetflixAmazonCredentials cred) {
	    // autowire beans
		super.accountAdded(cred)
		// enable agents etc.
	} 
	...
}
```

I was hoping to start small (easy migration to `Account` and `Credentials`) and refactors could be kept for later.

robzienert: > I was hoping to start small [...] and refactors could be kept for later.

This is a statement I've been seeing a lot more of lately in the community and I think we could do better. I appreciate the sentiment of making forward movement, but I have not seen frequent evidence that these smaller changes are followed up with a well-considered design and subsequent refactors. We definitely want to make incremental improvement and we definitely want to keep forward momentum; but more importantly, we want to make _forward-looking_, incremental improvement drawing on lessons we've already learned. We should be striving to make fundamental fixes with these proposals.

Account and credential management, especially in Clouddriver, have been long-identified as problematic and non-ergonomic, so I'm confused why we wouldn't want to tackle these problems immediately while we're in here, rather than copy-forward things we already know were mistakes. Doubly so, considering this code is a proposal for standardizing the way of representing account management across services, it's concerning we'd choose to implement the design knowing it isn't what we want.

Is there a path forward that we can think about what we want account and credential management to be, and work backwards from there to figure out what is possible in the short-term, and have a clear vision of where we're headed? I'd feel much more comfortable with incremental improvements when the destination is known, rather than incremental improvements without a known destination.

ncknt: > I was hoping to start small (easy migration to Account and Credentials) and refactors could be kept for later.

In the context of the comment, I meant implement the described design in smaller incremental change (by provider) - the goal of the document doesn't change.

I didn't feel the RFC's stated goals would be controversial but I'm okay revising them. I just don't understand what I'm missing. Adam had a good point (perhaps we don't need `Account` and `Credentials`) and I was proposing a way forward that remains manageable.

We're committed to implementing the change for Kubernetes and AWS providers (likely more providers and services) but we won't do all services and providers so we need an intermediary solution.

robzienert: I have no issues with the proposed goals, like I said earlier, I think this is a great idea and long overdue! I misinterpreted the context of your comment, and thought we were taking a step back. I apologize.

It is unclear to me what the actual proposed work is for the AWS provider, since that one is evidently the worst provider in terms of account mgmt. I see a bunch of work done for Kubernetes in your branch, but nothing for AWS - are we planning to rip out what is there and improve, or add this stuff on top?

rfc/account-management.md
ajordens: The `clouddriver-elasticsearch` module may not be all that relevant, I don't believe it's widely/if-at-all used outside of Netflix for entity tags. 

There is some lingering experimental indexing code that I suspect can be `rm -rf`'d ... years ago we experimented with maintaining a separate index in elasticsearch but it didn't go anywhere. Removing this code will likely simplify what was thought to be needed here.

rfc/account-management.md
ncknt: I just wanted to define the framework here and leave the discussion for each provider into the PRs that will ensue. I believe you (Netflix) have additional glue code around Spinnaker and I didn't want to break anyone's environment. But I'd be happy to simplify further.

ajordens: Things may get clearer in the draft PR if we can just be consistent about what we are calling things.

```
public interface CredentialsRepository<T extends Credentials> {
  T getOne(String key);
  Set<T> getAll();
  T save(String key, T account);
  T update(String key, T account);
  void delete(String key);
  String getType();
}
```

Minor point but potentially leading to confusion is the somewhat arbitrary use of `credentials` vs `account`. To someone not intimately familiar with the distinction, this will stand out. I think it's largely and easily correctable.

As a user of all this, I still don't think I should care about a distinction between Account and Credentials. Even more so looking across different Spinnaker services (`clouddriver` has made a mess of this but it's not as bad in igor right?).

rfc/account-management.md
```diff
@@ -0,0 +1,135 @@
...
+- extend the behavior of the `CredentialsRepository` across all types of accounts.
+- add a new account source and get the account information from its source of truth.
+
+We have evidence from our customers that some form of dynamic account management is needed.
+
+## Timeline
+
+Spinnaker 1.23 (1.22?):
+- Add interfaces to kork 
+- Start splitting AccountCredentialsRepository by provider (we'd now be able to use multiple accounts with the same name in different providers)
```
ajordens, line 49: Is that preferable? We've got a weird mix of APIs that sometimes include account + provider and sometimes just account.

Historically I've come down on the side of appreciating uniqueness in account names (but I could _possibly_ be swayed here).

rfc/account-management.md
ncknt: That is a good point. I actually changed the words in the middle of the RFC - hence the remaining `account` in the code snippet above. I used to have `Account` as `AccountProperties` (and `Credentials` and `Account`).

Would `AccountProperties` (or `AccountInformation`) be more explicit? Naming things is hard.

rfc/account-management.md
ncknt: I couldn't spot where the account was ambiguous if you know the provider (or the implied provider based on where you are in Deck or in the code). It clearly doesn't mean there aren't any!

I think it is important to allow different providers to use different sources. For instance, some users store their Kubernetes account information in a separate system but have a handful of manually configured AWS accounts that remain in Spring props.

The non uniqueness is a by-product of having separate sources of accounts. Currently, I am not even sure what happens when the same name is used in the different providers. The last one wins? And it probably changes if accounts are dynamic. Regardless, it's probably not what the user expects.

IMO we should make sure that the provider is always displayed next to the account (in Deck) or present in the URL in APIs.

ajordens: I didn't think `clouddriver` would even allow you to have duplicate account names across providers (we have two cloud providers that ultimately land resources in the same aws account but we give them distinct names.

ncknt: I believe you can configure it that way (in the yaml files) and I don't believe you'd get an error, just unexpected behavior.

ncknt:
One thing that I forgot to explicitly call out is that this is not just about creating accounts but also providing stable account management interfaces for users.

Yes we could simplify to: `class MyCredentials { // autowired props and beans }` but as soon as properties don't come from Spring, users will need to build `MyCredentials` objects and manually add the beans and there's no stability guarantee. I realize that this could also break if fields are removed from account definitions but that's very rarely happens.

That explains the `AccountParser` and `CredentialsLifecycleHandler` abstractions. In reality these are tiny interfaces.

I realize that there's also the issue of the AWS provider which has 5 credentials classes and a complicated loader mechanism. Ideally we'd keep that discussion for the AWS provider PR but for what it's worth, I share the concern that it doesn't look easily maintainable. I can also give it a try on that not-for-real PR if it helps clarify the end state.

ajordens:
Putting the prototypical PR aside, I'm a +1 with this proposal. Let's do it!

I think any concerns I've had are more related to potential implementations that can be tackled at that point.

robzienert:
+1 on the proposal.

PR #161, opened 2020-08-04
SIG registration for Oracle's OCI Platform

sandeeps2:


sandeeps2:
> * A link to the Slack channel for the SIG in the `spinnakerteam.slack.com` workspace

Will this be created for us. I am not a member of the `spinnakerteam.slack.com` yet.

ezimanyi:
> > * A link to the Slack channel for the SIG in the `spinnakerteam.slack.com` workspace
> 
> Will this be created for us. I am not a member of the `spinnakerteam.slack.com` yet.

You can go to https://join.spinnaker.io/ to create a Slack account; at that point you should be able to create a channel for the SIG.  General we've been choosing `#sig-{sig-name}` for consistency, so something like `#sig-oracle` or `#sig-oci` would be best.

sandeeps2:
> General we've been choosing `#sig-{sig-name}` for consistency, so something like `#sig-oracle` or `#sig-oci` would be best.

Thanks, I have created the #sig-oci channel and added it to the SIG README.

PR #162, opened 2020-08-04
feat(sig-as-code): add luispollo as lead, remove emjburns

emjburns:
Stepping down as SIG lead, and nominating @luispollo to replace me.

@spinnaker/sc PTAL


luispollo:
Thank you for the opportunity, @emjburns! ❤️

PR #169, opened 2020-08-26
feat(rfc): Amazon ECS end to end tests

allisaurus:
RFC which proposes Amazon ECS end to end tests + speculation on the future of alternatives.

### DIFF
--- a/rfc/ecs_e2e_tests.md
+++ b/rfc/ecs_e2e_tests.md
@@ -1,11 +1,11 @@
 # Amazon ECS End to End Test Proposal
 
-|               |                                                 |
-|---------------|-------------------------------------------------|
-| **Status**    | _**Proposed**, Accepted, Implemented, Obsolete_ |
-| **RFC #**     | _Update after PR has been made_                 |
-| **Author(s)** | Allison Stanko (@allisaurus)                    |
-| **SIG / WG**  | AWS SIG, Platform SIG                           |
+|               |                                                  |
+|---------------|--------------------------------------------------|
+| **Status**    | **Proposed**                                     |
+| **RFC #**     | https://github.com/spinnaker/governance/pull/169 |
+| **Author(s)** | Allison Stanko (@allisaurus)                     |
+| **SIG / WG**  | AWS SIG, Platform SIG                            |
 
 
 ## Overview
--- a/rfc/ecs_e2e_tests.md
+++ b/rfc/ecs_e2e_tests.md
@@ -1,0 +1,372 @@
+# Amazon ECS End to End Test Proposal
+
+|               |                                                 |
+|---------------|-------------------------------------------------|
+| **Status**    | _**Proposed**, Accepted, Implemented, Obsolete_ |
+| **RFC #**     | _Update after PR has been made_                 |
+| **Author(s)** | Allison Stanko (@allisaurus)                    |
+| **SIG / WG**  | AWS SIG, Platform SIG                           |
+
+
+## Overview
+
+This document proposes adding Amazon ECS provider tests to the Spinnaker 
+nightly builds in order to ensure breaking changes to the provider are 
+not merged into community release artifacts. 
+
+Since longer-term plans to retire the Jenkins-run end-to-end (e2e) tests 
+[are known](https://github.com/spinnaker/governance/pull/154/files#diff-856398ec244cbaddc659e4189e904a87R23-R26) 
+but not yet scoped/scheduled, this RFC presents a design for tests 
+built for the existing e2e setup, as well ideas about how these tests 
+could be mostly supplanted by integration tests or future monorepo CI workflows.
+
+
+### Goals and Non-Goals
+
+Goals:
+
+* Clarify the benefit of adding ECS end to end tests to the nightly builds
+* Outline high-level test design and requirements
+* Call out specific areas where more information is needed 
+
+Non-Goals:
+
+* Modifying the existing AWS e2e tests or porting them over to some future test process
+* Exhaustively testing every possible ECS server group configuration
+* Address Deck UI testing for the ECS provider (to be addressed separately)
+
+
+### Terms as used in this doc
+
+ - **e2e tests** : end-to-end tests; those that exercise all Spinnaker 
+   services involved in the deployment of a server group or other atomic operation.
+ - **integration tests** : tests that run in a single repo, against 
+   a live application (such as `clouddriver`), but do not require standing 
+   up a fully deployed Spinnaker installation. Ideally run as part of 
+   some automated workflow (e.g., pull request submission)
+
+
+## Motivation and Rationale
+
+Currently, the Spinnaker nightly builds run end to end tests that exercise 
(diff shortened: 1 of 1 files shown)


rfc/ecs_e2e_tests.md
```diff
@@ -0,0 +1,372 @@
...
+
+* Test psuedo code w/ annotations re: implementation questions is 
+available [here](https://github.com/allisaurus/buildtool/blob/ecs-test/testing/citest/tests/ecs_server_group_test.py), 
+for context.
+
+
+#### Test environment : AWS Account changes
+
+The ECS provider itself does not own any credentials, but points to a configured 
+AWS provider account to deploy to. The ECS provider can leverage an
```
allisaurus, line 193: Use of existing AWS test account based on conversations w/ @plumpy ; Ideally it would be managed by the CDF prior to implementation.

rfc/ecs_e2e_tests.md
```diff
@@ -0,0 +1,372 @@
...
+0. ASSUMES: load balancer, target group, and S3 bucket available in account or otherwise bootstrapped.
+1. Create a Spinnaker application with the AWS and ECS providers enabled.
+2. Create a server group (step also uploads JSON file to S3).
+3. Resize the server group.
+4. Destroy the server group (and the uploaded artifact).
+5. Delete the application.
+
+* Test psuedo code w/ annotations re: implementation questions is 
+available [here](https://github.com/allisaurus/buildtool/blob/ecs-test/testing/citest/tests/ecs_server_group_test.py), 
+for context.
```
robzienert, line 187: I would strongly prefer that E2E tests be written in Java or Kotlin.

citest is really difficult to understand and the person who originally authored and maintained it is no longer a part of the project. Even when it was actively maintained, we (Netflix) avoided using it because it was so cumbersome to write tests with. With this renewed interest in E2E, I'd like us to take a step back and evaluate something that is more likely to be in the wheelhouse of contributors to Spinnaker. Switching to the JVM will make it more likely that ownership and expertise will be distributed more broadly than one or two people.

Furthermore, whatever we land on, I want to strictly avoid Groovy, even for the test sourceset.

rfc/ecs_e2e_tests.md
```diff
@@ -0,0 +1,372 @@
...
+## Future Possibilities
+
+Big upheavals in how customers use the ECS provider or how the 
+community builds & tests releases will impact the longevity of 
+any end to end testing added today. Examples may include:
+
+* Elimination of all end to end testing prior to Spinnaker releases 
+  (assume then there would be an alternative that could be leveraged)
+* Core providers whittled down to excluded ECS or all cloud providers 
+  extracted to live as pluggins (would then leverage new system/guidance;
```
robzienert, line 367:
```suggestion
  extracted to live as plugins (would then leverage new system/guidance; 
```

rfc/ecs_e2e_tests.md
allisaurus: Understood re: wanting to move to an e2e stack that's more amenable to community maintenance, though for the sake of exercising ECS code paths in the near-term, I thought taking a crack at embellishing the existing stack was a reasonable interim step. 

But! If that's a non-starter: are there currently any JVM-based e2e tests running against the nightly builds, or other prior art to build off of? Happy to innovate here as well, but want to leverage what exists as much as is practical. 

If it's a total green field effort, then it sounds like a candidate for a dedicated RFC. 

> I want to strictly avoid Groovy, even for the test sourceset.

+1.  The (limited) integ test stuff I have working is in Java and anything we (ECS) contribute JVM-wise should be too.

rfc/ecs_e2e_tests.md
```diff
@@ -0,0 +1,371 @@
...
+
+### About the test design
+
+* Whether any Jenkins/e2e test environment configuration would be impacted 
+  by this change but is not immediately obvious given the [`buildtool`](https://github.com/spinnaker/buildtool) code.
+* Whether testing a provider that relies on another provider (e.g, AWS) 
+  might cause issues with an unknown (to me) authentication or test mechanism.
+* What's the best way to test any new configuration + test code prior to merging? 
+* Unknown/show-stopping restrictions on the AWS account currently used during tests.
+* How many regions are running the e2e tests, and would we need a unique ECR repository for each?
```
ezimanyi, line 309: I think they only run in one region, though that region comes from an environment variable.  It's currently set to `us-east-1` though there have been cases where we've moved things if a particular region was having issues and we didn't want the tests to fail.

rfc/ecs_e2e_tests.md
```diff
@@ -0,0 +1,371 @@
...
+within the test stack, the additional surface area is mostly contained 
+to the contents of the test itself and any ECS-specific knowledge 
+needed to debug failures. 
+
+* Additional resources must be created and maintained in hosting account
+    * Mitigation: Use an infrastructure as code mechanism to modify/recreate (template can live alongside test, maybe?)
+* Test flakiness and overhead for build cops & release managers
+    * Mitigation: Pre-merge testing, maybe non-blocking test runs prior to triggering build fails? (needs input from community)
+* Upstream/AWS-side API errors (already a risk for AWS provider). 
+    * Mitigation: Hopefully rare, but can be clearly identified by API responses/error msgs
```
ezimanyi, line 352: We definitely see this from time to time, but as you say this is usually reasonably easy to identify based on error messages.

rfc/ecs_e2e_tests.md
```diff
@@ -0,0 +1,371 @@
...
+      and is a resource typically created/killed during other Spinnaker e2e tests
+    * Could be created ahead of time in the same Cloudformation stack as 
+      the ECS cluster to minimize API calls during execution. 
+    * Could be created during the test via calls to the AWS provider, although 
+      this would duplicate executions of this code path during testing. 
+
+
+#### Test environment: Spinnaker Configuration
+
+To support ECS, the Spinnaker deployment under test would need to be configured with:
```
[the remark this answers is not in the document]
allisaurus: > One thing to consider as far as configuring Spinnaker, is whether to use an imperative series of hal commands or to **have a checked-in hal config up front and just run hal deploy apply** pointing at that.

This sounds so much better than running `hal` commands but yeah, wondering how we could accomplish this on the existing setup. I guess maybe the `ecs` and ECR docker registry stanzas could be wholesale added to the config? Testing how a mixed approach would work sounds daunting, but probably worth at least thinking about. Appreciate the suggestion!

rfc/ecs_e2e_tests.md
robzienert: Unfortunately, no - there are not any existing JVM-based integration tests running in the community. You're right that it would probably be worth an RFC in its own right.

allisaurus:
@plumpy @ajordens I plan to move forward with implementation soon. Can this RCF be considered "Accepted" for those purposes, or should I wait for you both to weigh in first?

ajordens (approved):

plumpy (approved):

PR #180, opened 2020-09-24
feat(sig): Add guidance for meetings

robzienert:


duftler (approved):
Looks reasonable to me. Out of curiosity, was the 'Meeting Recordings' section inspired by another project's SIG meeting guidance?

sig-guidance.md
```diff
@@ -0,0 +1,23 @@
+# SIG Guidance
+
+A collection of guidance for Special Interest Groups.
+
+## Meeting Notes
+
+Notes must be taken during SIG meetings and made available to the public, which are linked from the SIG's README.md file in the governance repository.
```
demobox, line 7:
```suggestion
Notes must be taken during SIG meetings and made available to the public, linked from the SIG's README.md file in the governance repository.
```

To avoid the potential confusion that it would be "the public" that would be linked

sig-guidance.md
```diff
@@ -0,0 +1,23 @@
...
+
+To maintain good cadence, it is generally recommended to have a scribe taking notes who is not the person facilitating the meeting or presenting.
+
+## Meeting Recordings
+
+All SIG meetings should be recorded.
```
demobox, line 13: Is this a "should" or a "must"? I.e. would it be acceptable for a SIG that has members that are uncomfortable with recordings to decide that they will _not_ record their meetings?

pstout: I think that it is important that notes be taken and publicized.  I disagree with the position that all meetings *must* be recorded, as I think it may discourage participation.  Watching a recording is fine if you are interested in consuming information.  It does not work if you want to provide input.

pstout (changes requested):
I like the guidance about note taking being required and standard locations for notes and recordings, when they are made.  I disagree with mandating that meetings be recorded.

sig-guidance.md
robzienert: Right now, I'm positioning it as a "should" - optional, but encouraged. 

A lot of the community is asking for recordings because they cannot participate due to timezones or other reasons. With this feedback, I'd like to move to mandatory recordings eventually, but would like to make that decision more deliberately once we've established whether or not it does indeed hinder participation, and if the community is getting use out of it.

robzienert:
@duftler I did not pull this language specifically from any other community, but I did take the general structure from how I understand the Kubernetes SIGs handle recordings (all SIG meetings are mandatory recording, advise people to leave if they do not provide consent).

sig-guidance.md
demobox: Could we emphasize that "should" means "optional", unless we think that'll be clear to readers?

Thanks for the clarification!

ajordens (approved):

markyjackson-taulia (approved):
non-binding +1
This is a great body of work for the community and I thank you for doing it

ajordens:
Nudge on this, are we good to approve the new guidance and begin rolling it out across SIGs?

sig-guidance.md
pstout: How about the following:

We encourage SIGs to record their meetings so they are available to people in other timezones.  At the same time, we understand that recording meetings may hinder participation in the meetings, so SIGs may choose not to record meetings in general or specific instances.  If a SIG chooses to not record a meeting, please inform the SC and TOC so that we have awareness of how often this is happening.  It is particularly important that notes are taken and published for any meeting that is not recorded.

pstout:
@ajordens @robzienert 
I suggested a change to the prose about recording meetings.  If that change, or something similar, is acceptable to you and others, then I am ready to approve this PR.

sig-guidance.md
robzienert: This sounds good, I'll update with this exact language.

pstout (approved):

PR #206, opened 2020-12-15
chore(core): Remove ezimanyi as TOC member

ezimanyi:
As previously announced, I'll be stepping away from the Spinnaker project and thus stepping down from the TOC effective today. This commit removes me from the various places where I am referenced as a member of the TOC.


plumpy (approved):
👍 but also 😭

robzienert (approved):

robzienert (approved):

PR #207, opened 2020-12-16
chore(core): Remove my username from one more template

ezimanyi:
I missed this in my last PR.


plumpy (approved):
I wonder if we can set assignees to `@spinnaker/toc` or something. That would be nice, but we can mess around with that later.

ezimanyi:
Yeah, I briefly looked at the documentation on templates to see if groups were allowed but it wasn't clear and I didn't want my last PR here to break them 😁

PR #215, opened 2021-02-03
Change SIG lead dorbin --> aimeeu

dorbin:


dorbin:
@aimeeu

brian-armory:
+1 to adding Aimee as the leader. Once we find a another suitable lead, I will cede my seat so that the leadership for the SIG isn't two people from one company.

aimeeu:
While Dave Dorbin can never truly be replaced, I'm happy to step into this SIG role. In 2019, I worked closely with a K9s Docs SIG Lead while I was under contract to the CNCF. Prior to that, I was the Docs Project Team Lead for Acumos (an LF project).  I'm currently a TW at Armory.

ajordens:
TOC +1

Seeing as how @aimeeu is onboard, I'll merge.

PR #227, opened 2021-03-08
Add bijnagte to netflix-new-hires

jonsie:


membership.yml
```diff
@@ -18,14 +18,12 @@ organization:
     netflix-new-hires:
       description: A no-permissions group for new Netflix hires such that they can be assigned as reviewers of PRs.
       members:
-      - gal-yardeni
-      - luispollo
-      - srekapalli
-      - anotherchrisberry
```
jonsie, line 24: gal, luis, and sairam are on the approvers list.  Moving anotherchrisberry from new-hires list to reviewers.

PR #246, opened 2021-06-22
chore(docs): Removing @robzienert from TOC

robzienert:
Removing myself from the TOC, as I've not been filling this role for basically the whole year.


emagana:
Thank you Rob for all your contributions! Spinnaker community wouldn't be the same without your leadership.

PR #256, opened 2021-08-25
add David Byron as platform sig lead

dbyron-sf:
I've been active in the platform sig for the past few months, and am happy to co-lead with @link108 .


plumpy:
Sounds great, thanks!

PR #268, opened 2021-09-16
security(rfc): Security scan proposals

jasonmcintosh:


mattgogerly:
So this sounds like a good idea to me - my question would be blocking PRs with vulnerabilities identified. Let's say hypothetically a new vulnerability is found which is already in master, would CodeQL block all PRs involving the relevant files until that is resolved?

jasonmcintosh:
I believe CodeQL will only block NEW PRs once it's added to branch protection rules:
* Nominally Enable code scanning
* DO NOT enable branch rules on the CodeQL findings that would prevent merge
* This should identify failures on the repos we could fix independently.

jasonmcintosh:
From discussion (will update PR):
* We may not do this on PR's but on a schedule and reports should be visible only to code maintainers initially.

PR #286, opened 2022-02-04
feat(rfc): user external secrets usage and delivery

jvz:
This RFC proposes new APIs for allowing users to bring their own secrets
from external secrets managers and use them in Spinnaker for various
purposes.

This idea made its public debut in Slack with early details in [this thread](https://spinnakerteam.slack.com/archives/CFN8F5UR2/p1643843327507219). Based on some initial feedback, I turned this into a full-fledged RFC as the idea is fairly cross-cutting across Spinnaker.

### DIFF
--- a/rfc/user-secrets.md
+++ b/rfc/user-secrets.md
@@ -101,7 +101,7 @@ Then using this secret in a new Clouddriver account:
 
 ```json
 {
-  "@type": "kubernetes",
+  "type": "kubernetes",
   "name": "eks",
   "permissions": {
     "READ": ["admin", "sre", "dev"],
@@ -109,7 +109,7 @@ Then using this secret in a new Clouddriver account:
   },
   "context": "eks",
   "namespaces": ["default", "sandbox"],
-  "kubeconfigFile": "secretFile://secrets-manager?r=us-east-1&s=arn:aws:secretsmanager:us-east-1:123455432100:secret:eks-sa&e=json&k=kubeconfig"
+  "kubeconfigContents": "secret://secrets-manager?r=us-east-1&s=arn:aws:secretsmanager:us-east-1:123455432100:secret:eks-sa&e=json&k=kubeconfig"
 }
 ```
 
--- a/rfc/user-secrets.md
+++ b/rfc/user-secrets.md
@@ -50,9 +50,11 @@ Further integrations may be developed beyond Q2 before a beta version is release
 
 ## Design
 
-A new `UserSecret` class is added to `kork-secrets` similar to the existing `EncryptedSecret` class.
-This class supports two new URI schemes: `secret://` and `secretFile://`.
+User secrets implement `UserSecret` which provides some common operations around extracting secret data and checking permissions.
+Implementations provide a `type` property which determines the structure of the secret data.
+These secrets can be referenced through a `secret://` (or `secretFile://`) URI using the `UserSecretReference` class to parse.
 Each URI is structured as `secret://secret-engine-id?param1=value1&param2=value2&param3=value3` where _secret-engine-id_ is the `SecretEngine` identifier and the parameter key/value pairs are specific to each engine.
+`SecretEngine` is updated with additional `decrypt()` and `validate()` methods for `UserSecretReference` values.
 Each engine is configured to make requests for secrets using a single Spinnaker identity, typically the workload identity of Spinnaker itself for ease of integration with cloud-specific APIs.
 To demonstrate, suppose we have a secret named `my-spinnaker-secret` in AWS Secrets Manager in the `us-west-2` region under the AWS account id `222211110000` while Spinnaker is running in EKS in the `us-east-1` region under the account id `111122223333`.
 This secret would have an [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) of `arn:aws:secretsmanager:us-west-2:222211110000:secret:my-spinnaker-secret`.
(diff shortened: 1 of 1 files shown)


rfc/user-secrets.md
```diff
@@ -0,0 +1,204 @@
...
+These secrets provide authorization metadata for integration with Fiat so that multiple tenants can share the same Spinnaker environment.
+
+### Goals and Non-Goals
+
+* Goals
+  - Enable Spinnaker to use external secrets managers through a common Spinnaker identity which secrets owners can grant read access.
+  - Enable Spinnaker to enforce access controls to these secrets based on authorization data delivered with the secret.
+  - Enable Clouddriver account credentials to use authorized user secrets.
+  - Enable SpEL expressions to use user secrets.
+  - Enable other Spinnaker services to use user secrets.
```
jvz, line 23: Remove pipeline-related goals of this RFC. Can be done in a follow up RFC.

rfc/user-secrets.md
```diff
@@ -0,0 +1,186 @@
...
+This needs to be designed so that secrets can be rotated while also avoiding unnecessary load on the relevant secrets manager APIs.
+* How to configure multiple authentication identities for use with fetching secrets.
+This currently relies on having a single Spinnaker identity to route secrets requests through, though in a more dynamic environment, there may be multiple workload identities in Spinnaker.
+* How this would work in other cloud providers.
+Kork only includes `SecretEngine` implementations for AWS and GCP; implementing equivalent engines for other cloud providers is out of scope for this initial RFC.
+
+## Security, Privacy, and Compliance
+
+This feature is a security feature and thus must be carefully reviewed.
+The proposed architecture involves various Spinnaker services being granted read access to secrets along with those services keeping cached secrets from multiple tenants.
```
link108, line 162: Not a blocker, but I could see this causing rate limiting issues

jvz: Potentially, yeah. I already updated my own custom `SecretEngine` implementation to configure some auto-retry with exponential backoff logic on API calls when being rate-limited. I'm thinking it could be useful to configure Spring caching on some of these beans, though that will require evaluating the security of encrypted cache stores like Redis for this purpose.

link108:
I like this idea a lot, it looks like it doesn't have any breaking changes with the existing secrets manager implementation. I do like the updated secret format, I'm curious on your thoughts re: standardizing on this format in the future, which *would* be a breaking change for secrets in the Spinnaker configuration

jvz:
I should note that I've been working on an updated version of this feature that doesn't require re-encoding secrets. Instead, the related metadata can be placed in external places such as secret tags in AWS Secrets Manager. That also makes it so that you can still use the same secret payload format that other AWS services assume (a string/string map encoded as JSON) which allows you to edit said secrets via the AWS console. The tag names are prefixed with `spinnaker:` to avoid name collisions with custom tags.

I'm also defining some basic cache config (via Caffeine) to help avoid hammering external secrets manager services. Overall, this latest revision is working out quite nicely so far. I'll have an RFC update within the next week or two I expect.

mattgogerly:
I do wonder if extending this functionality to allow arbitrary injection of secrets, e.g. into Kubernetes secret manifests, has merit. I know the initial scope of this RFC is for managing secrets for account definitions, but I think the Spinnaker project in general has a gap wrt deploying secrets.

jvz:
Yeah it would make sense in other contexts. I'm currently investigating how to integrate it into Kayenta for example.

rfc/user-secrets.md
```diff
@@ -0,0 +1,186 @@
...
+The alpha version of this feature is targeted for 2022Q2 and is being contributed by Apple.
+Development began in Q1 coinciding with the related Clouddriver Account Management API feature which initially integrated with the Kubernetes V2 provider.
+User secrets will be integrated into this API along with updates to existing Kork `SecretEngine` for AWS Secrets Manager for a reference implementation.
+
+## Design
+
+User secrets implement `UserSecret` which provides some common operations around extracting secret data and checking permissions.
+Implementations provide a `type` property which determines the structure of the secret data.
+These secrets can be referenced through a `secret://` URI using the `UserSecretReference` class to parse.
+Each URI is structured as `secret://secret-engine-id?param1=value1&param2=value2&param3=value3` where _secret-engine-id_ is the `SecretEngine` identifier and the parameter key/value pairs are specific to each engine.
```
jasonmcintosh, line 53: I'm wondering if we could switch to THIS format for secrets vs. the encrypted: syntax to standardize this a bit more.  Overall I like this format :) More sense as an official URI structure!

rfc/user-secrets.md
```diff
@@ -0,0 +1,186 @@
...
+Permission checks involving this list of roles must check that the set of target resource roles are all contained in this list.
+* `data`: a map of keys to base64-encoded (basic variant) binary data.
+* `stringData`: a map of keys to utf8-encoded string data.
+
+For example, the following is a snippet of a secret named `eks-sa` in `us-east-1` in the AWS account id `123455432100` encoded as JSON:
+
+```json
+{
+  "type": "opaque",
+  "roles": ["admin", "sre"],
```
jasonmcintosh, line 90: From a privilege escalation:  Is this an "OR" Or "AND" operation.  E.g. a user must be in ALL these groups to read this secret, or ONE of these groups.  That comes into delegation concerns mostly on who can update a secret, and the potential transitive impacts, particularly on account operations.  NORMALLY for example pipelines to update it you must be a member of ALL the roles in a pipeline.  i could see needing a similar constraint on secret access.  But... maybe not as secrets are a bit different in downstream impacts.

rfc/user-secrets.md
```diff
@@ -0,0 +1,186 @@
...
+This API is designed for single tenants and provides no access controls.
+
+Dedicated secrets managers such as [Vault](https://www.vaultproject.io) provide much more sophisticated secrets management and delivery methods, though it is a complex system to manage in addition to Spinnaker and the cloud infrastructure where it runs.
+While Vault could become another `SecretEngine` provider in Kork, requiring Vault when some users may already be using cloud-specific secrets managers or other secrets manager software would be too inflexible.
+
+It may be possible to store the Spinnaker authorization data for secrets in secrets metadata APIs, though the implementation details on how each secrets manager exposes metadata varies quite a bit.
+Since these access controls are not particularly sensitive, they don't _need_ to be encrypted with the secret, though this is the most direct way to deliver said access controls such that Fiat can understand them.
+
+This `UserSecretManager` API would replace the tentative use of `SecretManager` in the Clouddriver Account Management API.
```
jasonmcintosh, line 139: Is there a plan to replace all the existing secret engine classes with this newer api/format structure? E.g. treat ALL secrets as user secrets?

jasonmcintosh (approved):
Overall I think this is a good direction with well thought out ideas & concrete examples! There's definitely some security aspects on access controls on the secrets which would be KEY to implement, but i think having a better solution for secrets management on APIs is a great enhancement.  
Steps as I understand it with this secret system
* pre-create secrets and part of secret payload is how to control access.  
* Then can use API's to use those secrets, and part of that API validates the caller with the secret operation.  What permissions do things like clouddriver run as which might not HAVE an explicit role?  Is clouddriver an "Admin" and thus access to all resources? OR would you grant "role-clouddriver" to a secret allowing clouddriver access?
* Sounds like every secret is a JSON structure if I understand all of this, which includes the "roles" who can access the secret.  there's an interesting aspect on this of... some secrets are just a simple values or files.  Wondering how that would handle binary secret data (base64 encoded stuff?  Separate entry IN the secret or a file reference?)

rfc/user-secrets.md
jvz: Only have to be in one of the groups because we do time-of-use checks as well.

rfc/user-secrets.md
jvz: User secrets are more of an extension to external secrets. Maybe they can be unified somehow?

jvz:
> * Then can use API's to use those secrets, and part of that API validates the caller with the secret operation.  What permissions do things like clouddriver run as which might not HAVE an explicit role?  Is clouddriver an "Admin" and thus access to all resources? OR would you grant "role-clouddriver" to a secret allowing clouddriver access?

The external secret grants read access to Spinnaker. Spinnaker doesn't allow any users to do anything with it until a time of use check is performed. So yes, you'd grant cloud driver access to your secrets basically.

> * Sounds like every secret is a JSON structure if I understand all of this, which includes the "roles" who can access the secret.  there's an interesting aspect on this of... some secrets are just a simple values or files.  Wondering how that would handle binary secret data (base64 encoded stuff?  Separate entry IN the secret or a file reference?)

That's only the metadata. The secrets can be JSON, YAML, or CBOR right now.

jvz:
This feature has been merged a while ago. Not sure the point of leaving this open for comment anymore.

PR #299, opened 2022-06-20
RFC : Implementing CDEvents to Spinnaker using Java SDK

rjalander:
RFC proposal for implementing [CDEvents](https://cdevents.dev/) to Spinnaker using [Java SDK](https://github.com/cdevents/sdk-java), which is developed and maintained by the [CDEvents community](https://cdevents.dev/community/)


link108:
Thank you @rjalander for the RFC! I'll be taking a look and sharing this around to get some feedback 🙏

rjalander:
Thank you for lookin into this, could someone please share any feedback/suggestions.

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,106 @@
...
+| **Status**     | _**Proposed**, Accepted, Implemented, Obsolete_ |
+| **RFC #**      | https://github.com/spinnaker/governance/pull/299 |
+| **Author(s)**  | _Jalander Ramagiri (`@rjalander`)_|
+| **SIG / WG**   | CDF-CDEvents, CDF-SIGEvents |
+
+## Overview
+
+Many organisations use more than one CI/CD tools as part of their pipelines and most of the times the organisations implement their own glue code in order to make the tools work with each other.
+
+Implementing [CDEvents](https://cdevents.dev/) to Spinnaker enables Spinnaker to communicate with other CI/CD tools with in the ecosystem. [CDEvents](https://cdevents.dev/) is a common specification for Continuous Delivery events based on CloudEvents.
```
jasonmcintosh, line 15: I like the goal here! Having a standardized reactive system for CD is a great design goal!

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,106 @@
...
+Spinnaker can produce and consume [CDEvents](https://cdevents.dev/) using [Java SDK](https://github.com/cdevents/sdk-java) which is developed and maintained by the [CDEvents community](https://cdevents.dev/community/).
+
+### Goals and Non-Goals
+Goals:
+- Event-Driven Standardization with CI/CD tools like Keptn, Tekton, Jenkins, Spinnaker, etc..
```
jasonmcintosh, line 20: A question:  Has this integration been done in any of those tools?  Do we have a concrete implantation we could easily integrate with for integration tests/validation?

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,106 @@
...
+Proposing a design based on the PoC outcome.
+
+A `CDEventsController` will be created with the API endpoints to produce and consume `CDEvents` from `events-broker` with request mappings,
+
+`@RequestMapping(value = "cdevents/consume", method = RequestMethod.POST)`
+
+`@RequestMapping(value = "cdevents/produce", method = RequestMethod.POST)`
+
+
+An `events-broker` will be created to send and receive events by knative-eventing and the sample URL can be configured as below in the cluster.
```
jasonmcintosh, line 52: I THINK This is probably the most questionable for me?  E.g. Gate right now is literally an edge router system, sorta like spring cloud gateway (i think it'd be nice to get it MUCH more just a spring cloud gateway implementation).  Echo already has an event broker system so... more details on how this would work with-in that framework maybe would help?  What's the resilience plan?  How will failures be handled?  Keep events in the broker & requeue, etc?  Just need some more details ;)  AND ... no provider specific implementation.  I'd prefer an API abstraction.  E.g. "Configure echo to add events broker knative, or sqs/sns or amqp or jms or google or "

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,106 @@
...
+spec:
+  broker: events-broker
+  filter:
+    attributes:
+      type: dev.cdevents.artifact.published
+  subscriber:
+    uri: http://<spin-service>:<port>/cdevents/consume
+EOF 
+```
+Once the event is received by Spinnaker API endpoint `cdevents/consume`, further can be handled to run pipelines based on the configured event data in Spinnaker.
```
jasonmcintosh, line 75: Is the intention that the (and making an assumption that it'd be gate -> echo) endpoint would return in YAML?  Is this API defined/accepted?  What's the maintenance path on these APIs?  E.g. how often do these API's change, are there SDK's needing updated, dependency impacts?

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,106 @@
...
+## Known Unknowns
+
+The approach will be demonstrated with the Spinnaker Community using `CDEvents` PoC.
+
+## Operations
+Admin needs to configure Pipelines that should run on consuming the different event types from `events-broker`
+
+## Future Possibilities
+
+`CDEvents` open the doors to scalable and decoupled interoperability within the software supply chain and create the potential for greater visibility and measurability of cloud native continuous delivery workflows.
```
jasonmcintosh, line 106: SO overall like the ideas here, what I'd like is a bit more depth both on:
* Implementation, e.g. how it'll actually flow, who's going to help maintain it going forward (aka a SIG for this?), and some details
* Concrete other providers that we can reference for docs/implementation details/etc.
* Impacts long term and security details.  E.g. what permissions would consumers use?  How could they be restricted from getting ALL events to only some events?  Etc.?

rfc/cdevents-spinnaker.md
jasonmcintosh: Note - I think maybe that's the goal on the above, but the more we can explicitly state some of those design goals the better!

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,106 @@
...
+
+Many organisations use more than one CI/CD tools as part of their pipelines and most of the times the organisations implement their own glue code in order to make the tools work with each other.
+
+Implementing [CDEvents](https://cdevents.dev/) to Spinnaker enables Spinnaker to communicate with other CI/CD tools with in the ecosystem. [CDEvents](https://cdevents.dev/) is a common specification for Continuous Delivery events based on CloudEvents.
+Spinnaker can produce and consume [CDEvents](https://cdevents.dev/) using [Java SDK](https://github.com/cdevents/sdk-java) which is developed and maintained by the [CDEvents community](https://cdevents.dev/community/).
+
+### Goals and Non-Goals
+Goals:
+- Event-Driven Standardization with CI/CD tools like Keptn, Tekton, Jenkins, Spinnaker, etc..
+- Create an `events-broker` that is able to produce and consume `CDEvents`
```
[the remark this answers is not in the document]
jasonmcintosh: Looked at this again, and mentioned it but sharing here:  Echo's implementation is PROBABLY the easiest path forward on this.  

For an example of a new Consumer: https://github.com/spinnaker/echo/tree/master/echo-pubsub-aws
https://github.com/spinnaker/echo/blob/master/echo-pubsub-aws/src/main/resources/s3.jinja for the jinja template used to do translation.  

I'd SUGGEST a new "notifications" type, aka "echo-notifications-queue" sub module (OR all of this as a plugin) to allow adding new notification to a different destination.  OR https://github.com/spinnaker/echo/tree/master/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo add it here as a sub config as maybe a "queue" type... I could see some of that needing a better abstraction (aka the kork piece) for allowing different broker types.  Probably some more thinking along how this would best integrate...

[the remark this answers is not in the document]
rjalander: @jasonmcintosh it was nice meeting you in the Summit and discussing in-person about this integration.

One thing about the new consumer that you suggested to add like "echo-pubsub-knative".
But if we want not to have a provider specific implementation, do you suggest using an existing Webhook from Automated Trigger types or using an existing Pub/Sub System type(amazon/google) even for consuming a CDEvent message format (This we need to explore more on how to consume the CDEvent format using these types)

As we already discussed having Custom Webhook Stages for producing the CDEvent to a specific event-broker over HTTP POST or We can add new a notification type for producing CDEvent message to different brokers.

Please share your thoughts.

jasonmcintosh (approved):

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,149 @@
...
+
+Many organisations use more than one CI/CD tools as part of their pipelines and most of the times the organisations implement their own glue code in order to make the tools work with each other.
+
+Implementing [CDEvents](https://cdevents.dev/) to Spinnaker enables Spinnaker to communicate with other CI/CD tools with in the ecosystem. [CDEvents](https://cdevents.dev/) is a common specification for Continuous Delivery events based on CloudEvents.
+Spinnaker can produce and consume [CDEvents](https://cdevents.dev/) using [Java SDK](https://github.com/cdevents/sdk-java) which is developed and maintained by the [CDEvents community](https://cdevents.dev/community/).
+
+## Goals and Non-Goals
+Goals:
+- Event-Driven Standardization with CI/CD tools like Keptn, Tekton, Jenkins, Spinnaker, etc..
+- Spinnaker can able to consume and produce CDEvents from/to `events-broker`
```
xibz, line 21: Spinnaker as it sits has no concept of an event bus. This would require reworking the current architecture on how Spinnaker could handle these events. There's this notion of a context that houses all events that is collected at the Orca service. This means Orca would need to somehow organize and re-build the context as the each stage/task/operation context is usually propagated back to Orca and is built that way, but with the event bus, that seems a little bit more tricky.

rfc/cdevents-spinnaker.md
xibz: +1 on that. This is way too high level. We need to figure out how to ensure the context remains the same if we do decide on an event bus.

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,149 @@
...
+curl -v -d '{"id": "1234", "subject": "event"}' -X POST -H "Ce-Id: HelloSpinnaker" -H "Ce-Specversion: 1.0" -H "Ce-Type: dev.cdevents.service.deployed" -H "Ce-Source: not-sendoff" -H "Content-Type: application/json" "http://<events-broker-url>/default/events-broker"
+```
+
+### Dependencies
+
+The CloudEvent dependencies(io.cloudevents) are required to implement produce/consume events 
+
+## Drawbacks
+
+
```
xibz, line 130: I think we need to highlight that CdEvents is still very early in its life, and things are subject to change. That's a drawback we should highlight. Further, highlighting that if the eventbus does go down, it would bring down the whole system, as before, if a single service of spinnaker failed, like echo, things would still behave somewhat. While we can definitely architect the system in a way that makes the eventbus resilient. It's another drawback.

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,149 @@
...
+## Known Unknowns
+
+The approach will be demonstrated with the Spinnaker Community using `CDEvents` PoC.
+
+## Security, Privacy, and Compliance
+
+
+## Operations
+Admin needs to configure Pipelines that should run on consuming the different event types from `events-broker`
```
xibz, line 143: Can we go in depth? This is really not saying much. Can we have an example here showing the changes needed for a single API for example?

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,149 @@
...
+
+- Increase in traceability with in the ecosystem from commit to deployment of an artifact.
+
+- Operators can deploy Spinnaker with `CDEvents` to run multiple pipelines using different CI/CD tools in ecosystem.
+
+## Timeline
+Proof of concept with Spinnaker, Keptn and Tekton CI/CD tools developed to communicate with each other about occurrences and running pipelines from building an artifact to deploy.
+
+The CDEvents integration process will be achieved in two Goals:
+Goal-1 : Introducing CDEvents to Spinnaker using Spinnaker-Webhooks
```
xibz, line 43: I think we can split these goals even more. I think goal #1 is perfect, but goal number 2 is quite large. I would rather introduce a message bus a service at a time, rather than trying to solve the whole problem all at once. Goal 2 can just have sub goals of like enable pub/sub CDEvents for Rosco.

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,149 @@
...
+The CDEvents integration process will be achieved in two Goals:
+### Goal-1 : 
+Introducing `CDEvents` to Spinnaker using Spinnaker-Webhook:
+
+Using an existing Spinnaker Webhook `CDEvents` will be consumed/produced and Spinnaker can communicate with other CI/CD tools using CDEvents.
+
+### Consume CDEvents:
+
+A Source `cdevent` will be configured for Webhook type Automated triggers from Spinnaker dashboard.
+The URL `http://<HOST>:8184//webhooks/webhook/cdevent` formed will be placed as subscriber uri to the `events-broker` with specific event type to receive.
```
xibz, line 62: How do we envision a CI and CD system talking? Is it through a single event bus where every system would be talk to it or would each system, CI and CD, have their own bus?

A single event bus for every system is pretty scary tbh. And if we decide on multiple event buses, maybe it makes sense for webhooks to be peer to peer.

xibz (changes requested):

xibz:
Thanks @rjalander for taking the time to write this! Im super excited for CDEvents. I do think this needs to be fleshed out a little more, however, but definitely a good start!

rfc/cdevents-spinnaker.md
rjalander: Yes, will add the mentioned drawbacks

rfc/cdevents-spinnaker.md
rjalander: Thanks @xibz, for sharing your comments. 
As per this design goals this will be achieved by using an existing Webhook and Pub/Sub Automated triggers, Spinnaker can able to consume and produce CDEvents from/to `events-broker`

rfc/cdevents-spinnaker.md
rjalander: As per Goal 2, we are planning to use an existing Pub/Sub system types(google/amazon) and not any new services to listen to CDEvents and this configuration needs to specify on which event type the Pipeline should be triggered(cdevent-consume-pubsub.png),
And planning to create a new Notification type to produce CDEvents. Adding new CDEvents notification type can be handled as different part If this is quite large.

rfc/cdevents-spinnaker.md
xibz: Thanks @rjalander. I think what might help is a diagram with multiple systems, eg some CI service, some CD service, and illustrate where the event bus(es) sit(s). Once we have that diagram it'll remove our assumptions, and allow us to directly ask questions against it.

rfc/cdevents-spinnaker.md
rjalander: In the PoC we demonstrated using single knative events-broker used by every system with in the CI/CD ecosystem (https://youtu.be/92Txy-USHg4), 
I think we can have multiple event-brokers but those needs an infrastructure level configuration to pass the events between the different event-brokers used for CI/CD systems.

With this design proposal, webhook can be used for Knative like event-brokers  and send a CloudEvent format using HTTP Post request [cdevent-produce-webhook](https://github.com/spinnaker/governance/blob/0c57e8853b33bc6a85ab113c343397075ed1fed8/rfc/cdevent-produce-webhook.png)

afrittoli:
We had a chat about this with @xibz and @rjalander during the CDEvents WG, and we thought it could make sense to have at least one ad-hoc session to flesh this out. What do reviewers think, would that work or do you prefer to continue the discussion asynchronously on the PR?

rjalander:
Thank you @xibz for reviewing this proposal, I have update the PR based on comments. Can you please have a look at the PR when you get some time, and also as discussed please let me know when can we meet for an audio chat on this discussion further.

rfc/cdevents-spinnaker.md
afrittoli: We have a few integrations at POC level, and we are working in parallel to integrated CDEvents natively in other tools. In more details:

- Tekton natively supports sending and ingesting CloudEvents. We built an experimental version of that feature that uses CDEvents instead and we plan on migrating that to the latest version of CDEvents
- Jenkins also supports CloudEvents, but not CDEvents yet - there is a lot of interest in a CDEvents plugin for Jenkins though, Fidelity (a Jenkins user) is very keen on adopting CDEvents
- Keptn natively supports CloudEvents, even though part of their events are "imperative" which is not currently supported by CDEvents. We included Keptn in several POCs through external adapters between CDEvents and Keptn CloudEvents
- We are working with other tools communities as well, like Testkube, who are helping CDEvents improve test events
- We also had a POC with Flux and a POC with Argo thanks to @bradmccoydev

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Overview
+
+Many organisations use more than one CI/CD tools as part of their pipelines and most of the times the organisations implement their own glue code in order to make the tools work with each other.
+
+Implementing [CDEvents](https://cdevents.dev/) to Spinnaker enables Spinnaker to communicate with other CI/CD tools with in the ecosystem. [CDEvents](https://cdevents.dev/) is a common specification for Continuous Delivery events based on CloudEvents.
+Spinnaker can produce and consume [CDEvents](https://cdevents.dev/) using [Java SDK](https://github.com/cdevents/sdk-java) which is developed and maintained by the [CDEvents community](https://cdevents.dev/community/).
+
+## Goals and Non-Goals
+Goals:
```
afrittoli, line 19: I'm not so familiar with Spinnaker, so I don't know if Spinnaker already has a mechanism to associate incoming events with the execution of tasks in Spinnaker. 

If not, from my experience working on POCs with other tools I would say:
- first, implement producing events: this is usually simpler to do and it introduces benefits in terms of observability. Events can be collected and stored to derive devops metrics about number of deployments etc
- second, implement the ability to receive events: this can be more complex because you will need a way to allow users to associate incoming events with things that should happen in Spinnaker. 

If this exists in Spinnaker already, then receiving events shouldn't be more complex then sending :)

link108:
A few of us met and discussed this RFC, notes are here: https://docs.google.com/document/d/1rZ28d0ytTMBof2hwHvji4-aKJugUw3nRzFqQpEv5IQY/edit?usp=sharing

rjalander:
Thank you all for the meeting and sharing your comments on this RFC design.
I have Updated the design as per the last meeting, please review and we will schedule a follow-up meeting If required for the same.
Thank you.

xibz (changes requested):
@rjalander I like the changes. Most of my feedback is just grammar and formatting. Otherwise this looks good to me

rfc/cdevents-spinnaker.md
```diff
@@ -0,0 +1,293 @@
...
+
+- Operators can deploy Spinnaker with `CDEvents` to run multiple pipelines using different CI/CD tools with in ecosystem.
+
+## Timeline
+The CDEvents integration process will be achieved in two Goals:
+Goal-1: Introducing `CDEvents` to Spinnaker using existing infra(Pub/Sub Automated triggers and Webhook stage)
+
+Goal-2: Integrating `CDEvents` to Spinnaker by implementing a new Webhook type and a new Notification preference for `CDEvents`
+
+The final implementation aligning with `CDEvents` vocabulary specification will take from 4-6 weeks.
```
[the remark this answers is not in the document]
rjalander: I just referred from other RFC, removing this.

rjalander:
Thank you @xibz for reviewing this and sharing your comments.

rjalander:
Please review the latest changes as discussed, and let us know if any follow-up meeting is required to take this RFC to the next step.

link108 (approved):

dsanyika (approved):
It looks like a good start.

link108:
Looking forward to this, thank you @rjalander for putting this together! 🥳

PR #303, opened 2022-08-03
feat(process): add rfc for backport policy change

dogonthehorizon:
Amend the existing [backport policy][policy] to allow for new features or
backwards incompatible changes in behavior provided they are introduced with
opt-in feature flags. Details in the RFC.

[policy]: https://spinnaker.io/docs/community/contributing/code/releasing/#release-branch-patch-criteria

### DIFF
--- a/rfc/feature-flag-backports.md
+++ b/rfc/feature-flag-backports.md
@@ -3,7 +3,7 @@
 | | |
 |-|-|
 | **Status**     | _**Proposed**, Accepted, Implemented, Obsolete_ |
-| **RFC #**      | _Update after PR has been made_ |
+| **RFC #**      | [#303](https://github.com/spinnaker/governance/pull/303) |
 | **Author(s)**  | @dogonthehorizon (Fernando Freire) |
 | **SIG / WG**   | Platform |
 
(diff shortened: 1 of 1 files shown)


link108:
SGTM! +1

link108 (approved):

jvz:
Can we flesh out the feature flag idea a bit? There are libraries like Togglz which provide more sophisticated ways for managing feature flags and configuration for said flags. This general idea, though, sounds good as it's how I've been contributing things so far.

link108:
Thank you for calling out the risks!

link108:
@jvz happy to wait to merge this for a more fleshed out plan 👍

dogonthehorizon:
Hey @jvz thanks for the review, I do want to address feature flagging in general but would prefer for it to be out of scope for this policy change. As you say its how we're currently doing contributions, but the larger process for doing so could definitely still use some work. If you're ok with this approach I'd like to merge and follow-up with another RFC just on the feature flag portion :D

jvz:
No blockers here. We can break that into a separate RFC.

link108:
Thanks all for the reviews! Merging this ✅

PR #324, opened 2022-11-28
RFC: Re-order executions in waiting queue

rjalander:
RFC proposal for implementing Re-order executions in waiting queue,
https://github.com/spinnaker/spinnaker/issues/6735

### DIFF
--- a/rfc/reorder-waiting-executions.md
+++ b/rfc/reorder-waiting-executions.md
@@ -13,7 +13,7 @@ If a section is not applicable, remove it._
 
 ## Overview
 
-_Curreently the waiting execuitions of a pipelines runs in FIFO order. But due to certain project urgencies and release purpose, we quite often need to change the executions order of a pipeline._
+_Currently the waiting executions of a pipeline runs in FIFO order. But due to certain project urgencies and release purpose, we quite often need to change the executions order of a pipeline._
 
 _Implementing the Re-order functionality helps to run the waiting executions in the order the user wants_
 
@@ -32,13 +32,13 @@ _Currently it is not possible If the user wants to prioritize specific execution
 
 ## Timeline
 _The code for CLI command reorder implementation is almost ready for review. 
-Plannig to implement this feature in two Phases._
+Planning to implement this feature in two Phases._
 
-_Phase-1: completing CLI command implementaion and make this feature available for customers to use and work on feedback if any._
+_Phase-1: completing CLI command implementation and make this feature available for customers to use and work on feedback if any._
 
-_Phase-2: Implemeting GUI for reorder functionality by using the backend that is available from CLI reorder implementaion in Phase-1._
+_Phase-2: Implementing GUI for reorder functionality by using the backend that is available from CLI reorder implementation in Phase-1._
 
-_Proposed timeline for Phase-1 whould be 1-2 weeks as this feature is demoed for some users and got the feedback.
+_Proposed timeline for Phase-1 would be 1-2 weeks as this feature is demoed for some Spinnaker users and got the feedback.
 Phase-2 would be 3-4 weeks as there are some existing issues in the waiting executions area of Spinnaker dashboard that needs to be fixed._
 
 ## Design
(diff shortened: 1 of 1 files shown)


rfc/reorder-waiting-executions.md
```diff
@@ -0,0 +1,116 @@
...
+
+_New buttons will be added in Spinnaker dashboard(GUI) for each waiting execution to move the specific execution towards UP/Down or Top/Bottom based on the current position of the execution._
+
+_New Gate API endpoint will be created for external access_
+>>:8084/pipelines/<execution_id>/reorder?reorderAction=<UP/Down/Top/Bottom>
+
+_New Orca API endpoint will be created to access the same way_
+>>:8083/pipelines/<execution_id>/reorder?reorderAction=<UP/Down/Top/Bottom>
+
+_The main business logic of the feature lies in Spin Orca service, where the **CompoundExecutionOperator** will update the Redis repo using the existing class **RedisExecutionRepository** and update the pending executions to the user selected position in the Redis repo using the Class **RedisPendingExecutionService**._
```
jasonmcintosh, line 61: Note, would need to handle SQL as well if this moved forward.

rfc/reorder-waiting-executions.md
```diff
@@ -0,0 +1,116 @@
...
+### Dependencies
+
+_This proposal depends on existing services Spin CLI, Spin Deck, Spin Gate and Spin Orca._
+
+_No new dependencies are needed, existing systems will be sufficient to implement this feature_
+
+## Drawbacks
+
+_Currently there is no Re-order functionality of an execution_
+_Why should we **not** do this?_
```
jasonmcintosh, line 81: Some things that OFTEN hit:
* How do you know what the order SHOULD be - by doing it in FIFO you handle SHA's which may not have a particular order, but you REALLY want the latest kinda thing.
* E.g. Git triggers - it's possible that re-ordering would cause incorrect sha pulls on artifacts.  SO there IS a question in my mind of whether this should by default be visible?  It seems... rather specific to a very specific set of cases.  Docker triggers would be similar where unintentionally you could deploy older docker images.

jasonmcintosh (commented):
SO overall, no reason COULDN'T do this - mostly I wonder about the interest in this feature?  E.g. how common or how many people really want this - would love to hear from others on these kinds of things ;). AND a few big caution flags on re-ordering pipelines mostly around different times of artifact handling issues.

rfc/reorder-waiting-executions.md
rebalag: The pipeline design will have significant bearing on usage of triggers in the delivery workflow. If the Git commit id from the trigger is used in the pipeline and not head of the branch and uses the state of pipeline context then it is possible to have two pipeline executions deliver correctly even if order is changed. However, it makes sense that one would want the latest. 

There are couple of possibilities for reordering executions of the same pipeline
* The pipeline is used by triggers from multiple branches and build from main branch needs to be prioritized in the queue while others need to execute as well
* The pipeline is triggered by commits in same branch and multiple commits in the queue while the latest needs to be prioritized while others can be discarded

The first should use a different pipeline design and the second one should discard older executions in the queue as execution in different order can cause problems with identifying latest. This could then be an option of one execution at a time while discarding older execution in queue that has not yet started.

rfc/reorder-waiting-executions.md
rjalander: Thank you for reviewing this, Jason. Do we need this to be implemented for DualPendingExecutionService and InMemoryPendingExecutionService too.

rfc/reorder-waiting-executions.md
JamieChe: Hi, 
Our Spinnaker users have similar request to re-order the waiting executions. But I think one of the most important reason is the FIFO is not working as expected in certain situations. I used to submit a ticket for this issue https://github.com/spinnaker/spinnaker/issues/6373. 

Can we consider to fix it first?

JamieChe: Thanks @rjalander . Then it would be nice to also include this re-order feature . Our users will be happy to have it.

rjalander: One another scenario where Spinnaker users want to Re-order the waiting execution of same pipelines is,
- When the Pipeline has waiting executions from different source of triggers and If we want to prioritize the specific source of trigger to validate on priority for any business needs.

jasonmcintosh: DEFINITELY want the FIFO fix in first, then... more discussion on impacts/changes required for this, which aren't in my mind exactly "small" at the moment.  And this is MOSTLY tied to trigger impacts on executions and what data is pulled/used for triggers...

rfc/reorder-waiting-executions.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Prior Art and Alternatives
```
jvz, line 83: Have you looked at the `orca-qos` API? There may be a way to extend that to implement this functionality.

rjalander:
Yes as per https://github.com/spinnaker/spinnaker/issues/6735, we have "n" parent pipelines that are configured as Automated triggers in another Pipeline to run when each parent pipeline is successful.

And this Pipeline with triggers is currently running with FIFO and all other executions are waiting in queue.
At this stage user wants to re-order the waiting executions as per needs, but not sure If Parent pipeline can decide on which priority the waiting execution should run at this stage.
Please correct me If I am wrong.

rjalander:
@jasonmcintosh @xibz wanted to bring this discussion up, can we have a look at this feature again and bring RFC to the next stage.

jimmycasey:
> Yes as per [spinnaker/spinnaker#6735](https://github.com/spinnaker/spinnaker/issues/6735), we have "n" parent pipelines that are configured as Automated triggers in another Pipeline to run when each parent pipeline is successful.
> 
> And this Pipeline with triggers is currently running with FIFO and all other executions are waiting in queue. At this stage user wants to re-order the waiting executions as per needs, but not sure If Parent pipeline can decide on which priority the waiting execution should run at this stage. Please correct me If I am wrong.

This WoW make sense for us also where we know priority of item from the start and when this high priority item arrives, we want it to be executed next (put to top of queue).

PR #325, opened 2022-11-30
Create RFC for Kayenta RBAC

jvz:
Surprise: it's not just about Kayenta! This proposal for RBAC in Kayenta includes support for account management, a new Fiat backend based on Open Policy Agent, and a brand new security concept in Spinnaker: role-based access controls.

_But wait, doesn't Spinnaker already support RBAC?_ Yes, in the same way that a body of water supports an aircraft.

This also introduces the concept of a **role policy** to abstract the concept of a bundle of permissions linked to external roles to allow for simpler user management of access controls on the resources they own. This can be extended into Clouddriver, Front50, Igor, et al., in the future, though Kayenta's lack of RBAC made it the perfect place to begin.

### DIFF
--- a/rfc/rbac-ng.md
+++ b/rfc/rbac-ng.md
@@ -69,7 +69,7 @@ The Fiat API will be updated to support permission checks on these new resources
 
 Account permissions are most applicable to metric store accounts where only the `READ` and `WRITE` permissions apply.
 Granting read permission to a metric store account allows a user to fetch metrics by querying its `MetricsService` bean.
-Granting write permission to a metric store account allows a sid to manage permissions of the account, update the account configuration, and delete the account.
+Granting write permission to a metric store account allows a security identity (e.g. group/user) to manage permissions of the account, update the account configuration, and delete the account.
 Note that write permissions in this context are unrelated to writing metrics to a metrics store; metrics should be written to stores from the deployed applications being measured.
 Access controls will be defined in terms of role policies which are intended to replace direct use of permissions maps.
 
--- a/rfc/rbac-ng.md
+++ b/rfc/rbac-ng.md
@@ -1,0 +1,154 @@
+# RBAC for Kayenta and Spinnaker
+
+| | |
+|-|-|
+| **Status**     | _**Proposed**, Accepted, Implemented, Obsolete_ |
+| **RFC #**      | _Update after PR has been made_ |
+| **Author(s)**  | Matt Sicker <msicker@apple.com> |
+| **SIG / WG**   | sig-security |
+
+## Overview
+
+Kayenta is a canary analysis service used by Spinnaker and other projects.
+It does not implement any form of access controls to its APIs beyond the authentication requirements of any reverse proxy in front of it such as Gate.
+Normally, access controls would be enforced via Fiat, though this approach has not scaled well and duplicates permission storage with a polling mechanism.
+Fiat is designed to collect resource metadata from other Spinnaker services and provide permission evaluation in return.
+The primary information Fiat provides access to are external role information for users along with a cache of application permissions which are otherwise stored by Front50.
+
+Kayenta is intended to work as a standalone application, however, and makes a good case study for an improved model for _role-based access controls_ (aka _RBAC_) across Spinnaker.
+To do this, the concept of _role policies_ will be introduced as an abstraction to link together roles and permissions to define the security policy of a resource such as an account.
+[Open Policy Agent](https://www.openpolicyagent.org) (OPA) is a flexible policy engine commonly used for handling authorization, compliance, and other policy-related concerns by evaluating expressions on documents (structured schemaless data).
+Fiat provides a natural facade for externalizing security policy enforcement with an engine like OPA.
+Starting with Kayenta, OPA needs to be informed of applications, metric store accounts, external roles, and role policies.
+Metric store accounts and role policies will need backing storage to allow for runtime management.
+Account credential management APIs should be added that work with these new access controls and allow permissions to be defined and managed on accounts while being configured.
+This will make Kayenta more multi-tenant friendly and provide guidance on how to further enhance RBAC in Spinnaker.
+
+### Goals and Non-Goals
+
+Goals:
+- Add support for role policies in Fiat with Gate APIs
+- Add API for storing and managing account definitions in Kayenta to integrate with user secrets
+- Add Open Policy Agent as a storage and permission evaluator backend for Fiat
+- Add support for role policies to Kayenta metric store accounts
+- Update Front50 to push application permission changes to Fiat
+
+Non Goals:
+- Provide APIs for managing security policies directly via Rego or other potential policy backends
+
+## Motivation and Rationale
+
+Kayenta APIs are all provided fairly openly assuming an admin has configured it with usable account credentials.
+Account credentials can be configured for metric stores and object stores, the latter of which can also be used for configuration stores.
+Object stores are used for storing and retrieving metric query results which are each assigned random UUIDs upon creation.
+Configurations are encouraged to be shared and are viewable by all applications by default; as such, these are the only object stores that list objects from the store.
+It is metric stores where lack of access controls causes issues in multi-tenant environments as sources for metrics are typically guarded by team-specific credentials.
+All these accounts are defined as Spring beans through configuration files provided by the Spinnaker admin.
+
+Supporting self-service configuration of metric store accounts requires some form of access controls as metric stores may contain sensitive data.
+A self-service API should learn from the equivalent feature in Clouddriver how to integrate with user secrets and manage account definitions to provide a similar API for Kayenta.
+Users will then be able to configure Kayenta accounts at runtime rather than requiring a Spinnaker admin to update Kayenta's configuration each time.
+Multi-tenant usage of Kayenta will become feasible.
+
+## Timeline
+
+The initial Kayenta and dependent work is being developed during 2023Q1.
+Further integration with other Spinnaker microservices, role policies, and permission change tracking would be targeted for afterwards in 2023.
+
+## Design
+
+This consists of a few related security elements: authentication, authorization, and credentials management.
+Authentication is handled by Gate typically, so this will reuse the existing `X-Spinnaker-User` header for denoting the principal of the request.
+User details are tracked by Fiat, so this data will need to be mirrored into OPA as a new backend.
+Authorization is the more complex part as permission data is stored by the various microservices that own the resources.
+These checks are typically handled through the `PermissionEvaluator` interface from Spring Security which takes three arguments: the authenticated user, the resource, and the permission.
+Permission evaluation will be given the context of what roles the requesting user has, what permissions apply to applications, and what permissions apply to accounts.
+As applications are checked by multiple services, these permissions will be pushed by Front50 into Fiat so that any service can use application permissions.
+External role data can either be pushed from Gate on login (and when service accounts are saved) or included as part of the input data to Fiat for policy decisions depending on if it's locally available (like being included as JWT metadata or X.509 attributes).
+The Fiat API will be updated to support permission checks on these new resources along with supporting role policies.
+
+Account permissions are most applicable to metric store accounts where only the `READ` and `WRITE` permissions apply.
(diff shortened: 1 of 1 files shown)


rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+Credentials management will be implemented through a SQL provider for managing metric store account definitions along with integration with user secrets.
+When credentials definitions are parsed into account credentials, the parser can replace user secret URIs with their decrypted data if the calling user has permission to read them.
+These credentials can be combined with existing configuration file provided accounts.
+A REST API will be added for saving and deleting credentials with the former providing an API for managing the permissions of said credentials.
+
+### Dependencies
+
+Open Policy Agent would be a new dependency from outside the project.
+OPA is commonly used with a lot of the technology that Spinnaker integrates with.
+This is intended to replace most of what Fiat currently does by externalizing authorization decisions which can enable massive code simplifications to Fiat if fully embraced.
```
mattgogerly, line 85: Do we plan to bundle an OPA distribution as part of a Spinnaker install, for example using Halyard or Armory's operator? Or are we saying an installation of OPA will become a requirement for using Fiat?

Granted installing OPA on Kubernetes is trivial, but it's a extra hopp people would have to jump through.

rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+The Fiat API will be updated to support permission checks on these new resources along with supporting role policies.
+
+Account permissions are most applicable to metric store accounts where only the `READ` and `WRITE` permissions apply.
+Granting read permission to a metric store account allows a user to fetch metrics by querying its `MetricsService` bean.
+Granting write permission to a metric store account allows a sid to manage permissions of the account, update the account configuration, and delete the account.
+Note that write permissions in this context are unrelated to writing metrics to a metrics store; metrics should be written to stores from the deployed applications being measured.
+Access controls will be defined in terms of role policies which are intended to replace direct use of permissions maps.
+
+Credentials management will be implemented through a SQL provider for managing metric store account definitions along with integration with user secrets.
```
mattgogerly, line 76: Is there scope to include SQL provider support for canary configs too in this RFC?

mattgogerly (commented):
I really like the intention and direction of this. Fiat is a huge pain point for any scaling. Couple of questions that jump out immediately.

rfc/rbac-ng.md
mattgogerly: Nevermind I just found that in future plans :)

rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+User secrets have their own permissions attached during secret delivery that Spinnaker uses to perform access control checks upon use.
+
+## Operations
+
+Enabling security in Kayenta requires an additional SQL database.
+If this is the first SQL database being introduced into a Spinnaker or Kayenta environment, this adds a new system to operate.
+
+The introduction of Open Policy Agent support for a Fiat backend makes fairly large changes in how Fiat is operated today.
+Instead of configuring Fiat with URLs to poll, the individual services that have permission data will be configured to push to Fiat through a common Fiat SDK.
+This should significantly reduce the chattiness and space usage of Fiat's current synchronization and polling.
```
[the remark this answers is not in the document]
jvz: The idea would be to use Fiat as a facade to OPA, so other services won't really need to know about OPA at all. So yes, it'll be a new endpoint in Fiat. The point here is to make a transition possible and painless. 😀

rfc/rbac-ng.md
jvz: Yeah, I plan to update Halyard at least because this would likely be a sidecar for Fiat. I can also look at the other distribution mechanisms we have here since like you said, OPA is fairly straightforward to deploy.

rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+Authentication is handled by Gate typically, so this will reuse the existing `X-Spinnaker-User` header for denoting the principal of the request.
+User details are tracked by Fiat, so this data will need to be mirrored into OPA as a new backend.
+Authorization is the more complex part as permission data is stored by the various microservices that own the resources.
+These checks are typically handled through the `PermissionEvaluator` interface from Spring Security which takes three arguments: the authenticated user, the resource, and the permission.
+Permission evaluation will be given the context of what roles the requesting user has, what permissions apply to applications, and what permissions apply to accounts.
+As applications are checked by multiple services, these permissions will be pushed by Front50 into Fiat so that any service can use application permissions.
+External role data can either be pushed from Gate on login (and when service accounts are saved) or included as part of the input data to Fiat for policy decisions depending on if it's locally available (like being included as JWT metadata or X.509 attributes).
```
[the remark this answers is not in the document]
jvz: That's a good question! One approach we could use is to keep a persistent cache of this data to distribute to OPA as a policy bundle. Another option would be looking at options for a backing store for OPA. In a more integrated OPA setup, you might keep a centralized cache of your identity provider data, but that's basically punting the problem to another team.

Today, I'd expect a loss of that info only if your Redis or SQL database are unrecoverable.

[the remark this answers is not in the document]
rebalag: OPA uses external data for context free decisions - https://www.openpolicyagent.org/docs/latest/external-data/
This means that the caching and scale issues with Fiat may not be addressed by OPA as Fiat will cache the user groups and resource permissions information.  
Assuming that the model is to push the data to OPA for permission checks and not have OPA connect to DB to fetch info at the time of checks.
In the current system, the data is lost only when the Redis or SQL data is lost. So restarting the fiat does not lose the data.

rfc/rbac-ng.md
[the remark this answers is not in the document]
rebalag: I do like the idea of externalizing policy decisions making it standard way of adding new permissions.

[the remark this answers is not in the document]
jvz: Based on some initial prototyping, we still need durable storage for some things like role policies and any future abstractions, though the OPA integration will still rely on that. A lot of scaling issues in Fiat right now come from polling and unoptimized document models for making policy decisions.

rfc/rbac-ng.md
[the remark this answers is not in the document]
jvz: Commented on another spot about it, but the larger scaling issue comes from how Fiat collects data versus how it evaluates policies. Right now, it polls for data and offloads policy decisions to the individual services to give the worst of both worlds. I'd like to flip this to pushing data to Fiat along with externalizing policies to Fiat for reusability across services (and theoretical optimizations related to policy deduplication).

rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+
+Non Goals:
+- Provide APIs for managing security policies directly via Rego or other potential policy backends
+
+## Motivation and Rationale
+
+Kayenta APIs are all provided fairly openly assuming an admin has configured it with usable account credentials.
+Account credentials can be configured for metric stores and object stores, the latter of which can also be used for configuration stores.
+Object stores are used for storing and retrieving metric query results which are each assigned random UUIDs upon creation.
+Configurations are encouraged to be shared and are viewable by all applications by default; as such, these are the only object stores that list objects from the store.
```
dbyron-sf, line 44: Could use clarification on
> these are the only object stores that list objects from the store

rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Design
+
+This consists of a few related security elements: authentication, authorization, and credentials management.
+Authentication is handled by Gate typically, so this will reuse the existing `X-Spinnaker-User` header for denoting the principal of the request.
+User details are tracked by Fiat, so this data will need to be mirrored into OPA as a new backend.
+Authorization is the more complex part as permission data is stored by the various microservices that own the resources.
+These checks are typically handled through the `PermissionEvaluator` interface from Spring Security which takes three arguments: the authenticated user, the resource, and the permission.
+Permission evaluation will be given the context of what roles the requesting user has, what permissions apply to applications, and what permissions apply to accounts.
+As applications are checked by multiple services, these permissions will be pushed by Front50 into Fiat so that any service can use application permissions.
```
dbyron-sf, line 66: Does `will be pushed by Front50 into Fiat` mean a migration of existing perms from front50 data stores to fiat?

rfc/rbac-ng.md
```diff
@@ -0,0 +1,154 @@
...
+External role data can either be pushed from Gate on login (and when service accounts are saved) or included as part of the input data to Fiat for policy decisions depending on if it's locally available (like being included as JWT metadata or X.509 attributes).
+The Fiat API will be updated to support permission checks on these new resources along with supporting role policies.
+
+Account permissions are most applicable to metric store accounts where only the `READ` and `WRITE` permissions apply.
+Granting read permission to a metric store account allows a user to fetch metrics by querying its `MetricsService` bean.
+Granting write permission to a metric store account allows a sid to manage permissions of the account, update the account configuration, and delete the account.
```
dbyron-sf, line 72: what is `sid` short for here?

jvz: Security identity. Typically the group name but can be a username potentially.

rfc/rbac-ng.md
jvz: Front50 will keep ownership of the data; it will mirror that data into Fiat for its own policy evaluation usage (typically stored ephemerally).

rfc/rbac-ng.md
jvz: I mean the object keys are only enumerated for config stores, not object stores in general.

rfc/rbac-ng.md
dbyron-sf, line 137: Do you figure the `X-SPINNAKER_ACCOUNTS` could/would disappear as part of this?

rfc/rbac-ng.md
dbyron-sf: Perhaps
```suggestion
Granting write permission to a metric store account allows a security identity (e.g. group/user) to manage permissions of the account, update the account configuration, and delete the account.
```
?

rfc/rbac-ng.md
dbyron-sf: OK, I wonder if other folks get tripped up on the language here, or if it's just me.  I'm not coming up with a better alternative at the moment though...

rfc/rbac-ng.md
jvz: Definitely. Realistically, the only thing we need to pass around is the userid; the rest can be offloaded to Fiat (and its own policy evaluations). However, since you can still define permissions directly on resources, we'll likely continue loading the user's roles as granted authorities to help simplify local permission checks without having to move them to policies.

dbyron-sf: ❗ 🎉 Color me excited :)

rfc/rbac-ng.md
jvz: For permission data, instead of Fiat periodically polling all the other services for permissions, those services will be updated to push permission data updates to Fiat through a new data collection endpoint. Fiat can make an initial call to the various services to get a current snapshot of permissions when its own cache is empty (or to reset the cache) and subsequently listen for permissions updates over time. This will allow us to keep a fairly up to date but eventually consistent cache of Front50 resource permissions, Clouddriver resource permissions, and any other services (Igor and Kayenta come to mind; maybe even Echo) with a couple advantages over the existing approach:

1. By listening to what is effectively an event stream, we keep a more up to date view of permissions than periodic polling.
2. By reducing polling to cold starts, we significantly reduce the network traffic Fiat generates.

A corresponding update to `FiatPermissionEvaluator` will allow for asking Fiat to do policy evaluation.

rfc/rbac-ng.md
jvz: Those are the three account types in Kayenta, though that itself is an odd design choice. What I'm getting at here is that it seems as though the metric store accounts are the most important to provide access controls on, but the object stores might also benefit depending on how sensitive your metrics are.

jvz:
This has rescoped a bit since I began. It seems like using OPA may not be necessary (at least not at first). I'll file a new RFC with the updated version later (if there's any reason to file RFCs anymore that is).

PR #336, opened 2023-01-31
RFC: Monorepo

jcavanagh:
This RFC proposes to consolidate all Spinnaker code into a single Git repository.

It also contains samples and instructions on how to preserve third-party forks, and maintain those forks moving forward.

### DIFF
--- a/rfc/monorepo.md
+++ b/rfc/monorepo.md
@@ -1,0 +1,557 @@
+# Monorepo, Composite Build, and Unified Test/Release
+
+|               |                                                             |
+| ------------- |-------------------------------------------------------------|
+| **Status**    | Proposed                                                    |
+| **RFC #**     |     |
+| **Author(s)** | Joe Cavanagh ([`@jcavanagh`](https://github.com/jcavanagh)) |
+| **SIG / WG**  | Platform SIG                                                |
+
+## Overview
+
+This document describes the unification of all Spinnaker service, UI, and directly supporting code into a Git monorepo.
+
+Within that repository, create a single build, test, and release process.
+
+This process will be somewhat intelligent - detecting changes in order to minimize the amount of work performed for a given PR or release build.
+
+This is closely related to [previously-closed RFC 154](https://github.com/spinnaker/governance/pull/154)
+
+### Goals and Non-Goals
+
+This document describes:
+
+- Consolidation of all Spinnaker service, UI, and directly supporting code into a single Git monorepo
+- Unified build, test, and release processes for that single repository, smart enough to minimize work for a given task
+- A mechanism to achieve this while also not destroying externally-maintained forks/extensions
+  - Methods for maintaining external forks - both of old single repositories, and the new monorepo
+- Date-oriented release versioning for all Spinnaker components, containers, and published libraries
+
+## Motivation and Rationale
+
+The many repositories that constitute Spinnaker today create excessive complexity and toil.  Build and release tooling is duplicated across many repositories, the release process itself is convoluted and laborious, and making changes to shared code is toilsome and error-prone.
+
+A monorepo significantly streamlines the contribution, build, test, publish, and release process, which is the source of the majority of the aforementioned toil.
+
+A major goal of the project is to improve new developer accessiblity.  A single repository to checkout, build and test a change, and submit a pull request is about the minimum possible burden to place on a new contributor.
+
+Spinnaker's microservices have significant shared code.  For example, he `kork` library is used by every other service, including handling the key role of defining dependency versions across the application.  Changes here take many steps and pull requests to propagate across the application - the history is cluttered with auto-merges of `kork` bumps, and breaking changes downstream are difficult to catch early.  The `fiat` libraries parallel `kork` in most services, and are also widely shared and often version-bumped.
+
+Spinnaker's microservices also share a common build target (the JVM) and share common build tooling (Gradle).  Gradle composite builds are an extremely natural fit for this project.
+
+### A Developer Workflow Example
+
+Developing on Spinnaker across its many repositories can be an unpleasant experience.
+
+This is an example workflow of how a developer would make a breaking change to a common library, `kork`:
+
+**The Separate-Repo Flow**
+
+1. Checkout `spinnaker/kork` and make the change
+2. Publish the `kork` library to `mavenLocal`
+3. Find all references to that class or method
+    - Let's say it's used in `orca`, `clouddriver`, and `fiat` for this hypothetical
+4. Checkout `spinnaker/orca`, `spinnaker/clouddriver`, and `spinnaker/fiat`
+5. Configure each of their Gradle builds to use the `kork` library you just published to `mavenLocal`
+6. Make the requisite changes in the `orca`, `clouddriver`, and `fiat` projects, and fix any tests/other issues that may arise
+7. Undo the Gradle changes to each project, removing the `mavenLocal` version of `kork`
+
+Unfortunately, we are not done yet - we still need to submit and integrate our four PRs in the proper order.
+
+8. Submit four PRs to each of our four impacted repositories
+9. The PRs to `orca`, `clouddriver`, and `fiat` will fail, since they do not have access to that new `kork` version
+10. Merge the `kork` PR, and wait for the new version to publish
+11. Update all three of the `orca`, `clouddriver`, and `fiat` PRs with the new `kork` pin
+    - If more than one breaking `kork` change is happening simultaneously, this change may cause conflicts on the `korkVersion` line
+12. Merge all three service PRs
+
+This is a significant amount of unnecessary toil - the level of effort required to make cross-cutting changes discourages new developers, and burns out maintainers.
+
+This inertia against cross-cutting change affects the quality of that code.  If it is not easy to change, it will remain unchanged, be shimmed or facade-ed, or other means engaged in order to avoid the above toil.
+
+**The Monorepo Flow**
+
+1. Checkout `spinnaker/spinnaker`, and make the change in the `kork` subtree
+2. The compiler will tell you what you have broken in `orca`, `clouddriver`, and `fiat`
+    - Since our IDE is now aware of all code in the project, we can find all references to that code in seconds, and have access to things like IntelliJ's automatic refactoring tools
+3. Make the requisite changes in the `orca`, `clouddriver`, and `fiat` projects, and fix any tests/other issues that may arise
+4. Submit a single PR containing the complete change
+5. Merge the single PR
+
+A single repository for all code has many benefits:
+
+ - No more version pinning or bumping
+ - Cross-cutting changes across several services can be done atomically
+ - Breaking changes in internal dependencies can be found quickly and resolved immediately
+ - Minimizes barriers to new contribution
+
+Expanding on the scope of the previous related RFC, `deck` and `spin` will also be moved into this monorepo.
+
+## Timeline
+
+TBD
+
+## Design
+
+### Gradle Composite Builds
+
+The key component of the monorepo design is [Gradle's composite build feature](https://docs.gradle.org/current/userguide/composite_builds.html).  Explained in short:
+
+1. All existing Spinnaker projects are combined into one source tree
+1. A top-level Gradle build is created, which "includes" all existing Gradle projects into one build context
+    - This requires little to no changes to the original projects.  Gradle will automatically understand Java conventions and dependencies, and rewrite libraries to use a just-built version from the local repository
+    - e.g. If a dependency of `implementation "io.spinnaker.fiat:fiat-api"` exists in any other service, Gradle will match that to an included build's Java-convention group/package name, build Fiat locally, and provide that output as the dependency to the downstream project
+    - One can imagine this as a completely automated `mavenLocal` -style workflow, but with the full power of Gradle behind it
+1. New top-level Gradle tasks can wrap these projects together in arbitrary ways, and allows us to implement a fully unified build, test, publish, and release process
+1. Consolidate all existing Github Actions tooling to leverage these composite tasks to build/test/publish/release one, several, or all services/libraries/etc
+1. New contributions can implement complex features across many services trivially, in a single pull request
+    - This drastically reduces the amount of labor required of both contributors and reviewers/committers
+    - A change that does not function as intended can be reverted (or fixed) quickly in a single pull request
+
+This does not come entirely for free, however:
+
+1. The build can be heavy.  There's a lot of graph for Gradle to chew through, and it can take a few minutes even with a "hot" Gradle daemon.  Cold startup is a few minutes longer, to get dependencies built once.
+    - Fortunately, Gradle caches work perfectly well with composite builds, and the cache has a *dramatic* impact.  Operating a Gradle cache is mandatory for CI (if the Gradle-GHA cache functionality is limited or insufficient), and should also be strongly considered for all development teams.
+    - It is notable that included builds can be referenced just like Gradle project syntax from the command line to drill down to specific tasks.  e.g. `./gradlew :fiat:fiat-api:test` is perfectly valid in composite-monorepo form.  This is very useful when making a smaller change that does not require waiting for a larger test suite to run locally, and Gradle is not building more than it needs to.
+1. With heavy builds comes longer CI times.  Some intelligence will need to be added to the Github Actions workflows to select Gradle tasks intelligently based on the incoming diff.
+    - We use this kind of functionality heavily in our own forks/extensions - in combination with the cache, our CI times are ~30 minutes for a full build and test of all Spinnaker code.  Code that does not touch `kork` can omit many tests, and takes between 5-15 minutes.
+1. Libraries still need to be published, even though the project itself no longer uses them - there are many external consumers.
+
+### Monorepo Github Location
+
+The monorepo will live at `github.com/spinnaker/spinnaker`.  All other repositories will be archived once the transition is completed.
+
+### Primary/Default Branch Name
+
+The primary/default branch of the monorepo will be renamed from `master` to `main`.
+
+### Yes, Deck Too
+
+This design explicitly integrates `deck` and `deck-kayenta` into the monorepo.  Cross-cutting changes do occur between the UI and the backend, and Deck is already a [Lerna](https://lerna.js.org) monorepo within itself - building all of it from its existing Gradle entry point is simple.
+
+Linking the builds of `deck`, and `deck-kayenta` is not something Gradle can do, but we have two good options here:
+
+1. Use the local `deck-kayenta` code during the build process of `deck`, using `yarn`'s workspaces feature
+1. Just integrate `deck-kayenta` as a regular `deck` module, like all the other Lerna subprojects under `packages/*`
+
+### The `spin` CLI
+
+This project is actually the most difficult to integrate into the monorepo, as golang builds tend to be really finicky about their environment.  I think Github Actions should afford us the flexibility to partition that out reasonably, but that requires some investigative work.
+
+There are definitely concrete benefits to including it in the build cycle - it can directly benefit from using adjacent Gate libraries, and it's much more testable when colocated against the APIs it is calling.  Setting up ephemeral CLI testing environments (on every change to Gate, even), is well within GHA's capabilities.
+
+### Halyard
+
+The inclusion of Halyard into the monorepo merits some discussion.  It is a tool everyone wants to see replaced, but has no replacement yet.  It consumes many Spinnaker service libraries, and would greatly benefit from composite builds.
+
+However, it sits in the unique position of not really needing to be up to date with its Spinnaker dependencies.  We need to weigh the benefits of building Halyard more conveniently against failing pull request builds because a contributor broke something in Halyard.  We should also consider that Halyard being stale gets harder and harder to fix over time, so perhaps the possibility of disruption to pull request builds is worth it in the long run.
+
+A middle ground does exist, where we can deactivate Gradle's composite subsitution for one or more Halyard dependencies if needed, and pin a previously published version of the library instead.  This at least leaves the possibility for a better Halyard build (at least one dependency unpinned), and the status quo (everything pinned) is the lower bound.
+
+### Integrating Gradle Build Plugins
+
+Gradle composite can also build plugins to be used in the same build, allowing us to also combine [Spinnaker's Gradle build plugin](https://github.com/spinnaker/spinnaker-gradle-project) into the project itself.
+
+This is the only non-Spinnaker-service version that was pinned in Spinnaker Java projects, and including `spinnaker-gradle-project` into the monorepo enables trivial iteration on the Gradle build process.
+
+### Releases and Versioning
+
+Currently, each Java project within Spinnaker maintains its own semver-shaped version designation pinned within various other Spinnaker projects.  This is unnecessary with composite builds, as the project builds as one cohesive unit without any dependency versions at all.
+
+However, containers, libraries, and other artifacts must still be published, and there is a need for a contract with library consumers (such as plugins).
+
+#### Spinnaker Release Versioning and Scope
+
+Spinnaker releases will use a date-based versioning scheme of `<year>.<number>.<patch>`, where `number` represents the number of major releases in that calendar year, and `patch` designates new builds produced for reasons outlined below.  The first numbered release of 2023 would therefore be `2023.0`, followed by `2023.1` and so on.  Patches to numbered releases are appended to the numbered release, e.g. `2023.0.1`.
+
+The initial version of a numbered release is considered to have patch version zero, even if not explicitly stated for brevity.  Unqualified numbered releases refer to the latest patch of that numbered release.
+
+Significant new feature work should result in a new release number.  Bugfixes, security fixes, or other important non-breaking, non-feature changes may be introduced in patch releases.  Small updates to existing functionality may be considered for patch releases if the changes are additive and non-breaking - this is left to the judgement of the reviewer and approver, along with the release manager.
+
+All patch releases (`<year>.<number>.<patch>`) should be fully compatible with that version's initial release, and all previous patches to that release number.  Patch releases should be fully non-breaking, from external API all the way down to public function signatures in libraries.  Exceptions can be made to mitigate critical bugs or security issues.
+
+Features and API annotated with `@Alpha` or `@Beta` annotations may change at any time.  However, it is desirable to minimize breakage in patch releases when possible, particularly if annotated as `@Beta`.  This is also left to the judgement of the reviewer, approver, and release manager.
+
+No contract is given between numbered releases (`<year>.<number>`) - these may break, add, remove, or otherwise change any functionality deemed necessary by the project.
+
+#### Release Timing and Cadence
+
+Release timing and cadence is not addressed by this RFC.  When a new release is made during a calendar year, the release number increases by one.  When a patch release is made, the patch number increases by one.
+
+The release manager retains the authority to release as they see fit.  Backports to older versions will be judgements made based on the criticality of the issue, risk to a release's stability, and the difficulty of backporting the change itself.
+
+#### Release Branch Names
+
+Release branches will have a naming scheme similar to today, but date-oriented.  A release branch is prefixed by `release-`, appends the release version, and is suffixed with `.x`, e.g. `release-2020.3.x`.
+
+This branch will produce containers, libraries, and other artifacts for that numbered release.
+
+#### Artifact Versioning Scheme
+
+Spinnaker library and container versions from release branches will mirror the release version (e.g. `2023.3.4`).  This should be defined in the root `gradle.properties` file, and used throughout included builds and subprojects to publish their respective artifacts.
+
+From `main`, library and container versions will be published as `main-<gha_run_id>`, e.g. `main-4041166540`.  Github Actions run IDs always increase, so this functions identically to timestamp-based snapshot versions with two key benefits:
+
+1. The workflow that produced the artifact can be located trivially, which is extremely helpful when diagnosing many build and publish issues.
+    - A Docker container named `main-4041166540` or jar named `clouddriver-aws-main-4041166540` can be found at `https://github.com/spinnaker/spinnaker/actions/runs/4041166540`
+1. The shorter build ID is a lot more human readable, and there is no need for.
+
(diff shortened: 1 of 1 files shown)


mattgogerly:
I am very much in favour of this. Couple of questions having skim-read the proposal.

- are we expecting the build process to require more resources (notably memory)? Some of the projects already require increasing the heap space to build
- are any changes required to plugins in spinnaker-gradle-project to support the monorepo?
- is Keel in scope to be included in the monorepo? It's quite the ugly duckling at the moment - I don't know anybody outside of Netflix (and maybe AirBnB?) using it. There's been a lot of issues getting Kork changes to merge.
- are you proposing a timeline for when we should cut over to the monorepo? After a specific release?

jcavanagh:
Thanks @mattgogerly !  Appreciate the feedback, and I've answered your questions below:

 > are we expecting the build process to require more resources (notably memory)? Some of the projects already require increasing the heap space to build

It is definitely a bit heavier, but not too excessive, I think.  Our monorepo build runs with `org.gradle.jvmargs=-Xmx4096m` and has been happy with that for a long time.  We do not currently build Halyard or Keel in that monorepo, which may bump up the requirements a small amount when/if those are integrated into an OSS monorepo.

As mentioned, using a Gradle cache (either in GHA or for local development) will dramatically improve memory usage and build time.

> are any changes required to plugins in spinnaker-gradle-project to support the monorepo?

I don't think it should require any significant changes - we use a wrapped version of that plugin in our monorepo today, excluding some of its parts internally like code formatting, etc.  That plugin will continue to be applied to each subtree's `build.gradle` as it is today -  Gradle composites consider each `includedBuild` a completely independent Gradle project, so the scope of that plugin's operation isn't actually changing.  

Since the `spinnaker-gradle-project` plugin can now be directly included in the project and also built compositely, changes made there can be tested immediately and always fit the current needs of the project - another pin we can remove.  When included in the project, it can and should still be published as a consumable plugin jar, as the current separate repo does - I am certain that library has many external consumers.  

> is Keel in scope to be included in the monorepo? It's quite the ugly duckling at the moment - I don't know anybody outside of Netflix (and maybe AirBnB?) using it. There's been a lot of issues getting Kork changes to merge.

I would suggest yes, as leaving it outside would only make its situation worse.  Halyard has similar problems, and leaving it out would also only degrade its quality further.  

Even when using Gradle composites, it is possible to pin a subtree to a previously-published library version if that is needed to unbreak a build.  I would say that is not generally advisable, and fixing every component should be part of integrating a breaking change.  

I would hazard a guess that the author of whatever change broke Keel had no idea it would break that, which is something that monorepo CI would immediately see.  There will certainly be some work to pay down existing technical library-pin debt, but I would expect that debt to accrue much more slowly once shared libraries can build against the entire project at once. 
 The author can either fix or make their change compatible once they know it's even breaking something at all, with all that context top of mind.  

> are you proposing a timeline for when we should cut over to the monorepo? After a specific release?

No timeline at the moment - once there is consensus on the approach, I would like to proceed to a POC to ensure it will work as advertised, and third-party fork maintainers can verify that they can still consume OSS changes and upstream their own changes as described in the document.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+
+There are definitely concrete benefits to including it in the build cycle - it can directly benefit from using adjacent Gate libraries, and it's much more testable when colocated against the APIs it is calling.  Setting up ephemeral CLI testing environments (on every change to Gate, even), is well within GHA's capabilities.
+
+### Halyard
+
+The inclusion of Halyard into the monorepo merits some discussion.  It is a tool everyone wants to see replaced, but has no replacement yet.  It consumes many Spinnaker service libraries, and would greatly benefit from composite builds.
+
+However, it sits in the unique position of not really needing to be up to date with its Spinnaker dependencies.  We need to weigh the benefits of building Halyard more conveniently against failing pull request builds because a contributor broke something in Halyard.  We should also consider that Halyard being stale gets harder and harder to fix over time, so perhaps the possibility of disruption to pull request builds is worth it in the long run.
```
dbyron-sf, line 147: I'd like to include halyard in the monorepo.  Catching breakage during e.g. spring boot upgrades (and gradle upgrades, java, etc.) would be super helpful.  It's true that it doesn't need to be up to date, but it may as well be.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+git subtree add -P fiat git@github.com:spinnaker/fiat.git master
+git subtree add -P front50 git@github.com:spinnaker/front50.git master
+git subtree add -P gate git@github.com:spinnaker/gate.git master
+git subtree add -P halyard git@github.com:spinnaker/halyard.git master
+git subtree add -P igor git@github.com:spinnaker/igor.git master
+git subtree add -P kayenta git@github.com:spinnaker/kayenta.git master
+git subtree add -P keel git@github.com:spinnaker/keel.git master
+git subtree add -P kork git@github.com:spinnaker/kork.git master
+git subtree add -P orca git@github.com:spinnaker/orca.git master
+git subtree add -P rosco git@github.com:spinnaker/rosco.git master
```
dbyron-sf, line 268: [spinnaker-monitoring](https://github.com/spinnaker/spinnaker-monitoring) is [below](https://github.com/spinnaker/governance/pull/336/files#diff-a56c8a7010c73454be672ac2b81d8df1e145318115a9a833fa9570da4b1dc36cR329), but not here....It's still in [the bom](https://github.com/spinnaker/buildtool/blob/f7c125019e4b7e67e9bf7f182ae4963fc854fdf9/dev/buildtool/bom_scm.py#L36), and not technically archived just yet.  Worth some discussion?

dbyron-sf:
I too am super excited to see this happen.  There are not enough emojis in the world for me to communicate this.

It doesn't build with gradle, and I haven't thought this all the way through, but could it make sense to include https://github.com/spinnaker/buildtool in the monorepo?  Perhaps it makes sense to include some detail here about release mechanics with a monorepo.  The current process is documented [here](https://spinnaker.io/docs/releases/release-manager-runbook/), but it basically involves tagging the HEAD of the relevant release branch in each repo and then running a [buildtool github action](https://github.com/spinnaker/buildtool/blob/f7c125019e4b7e67e9bf7f182ae4963fc854fdf9/.github/workflows/release.yml).

Maybe the release process with a monorepo will still involve pushing a tag to release artifacts (i.e. docker containers, debian packages, jars)?  And maybe we'll still use buildtool to publish boms, etc.?  So though buildtool will need some adjustment, maybe the general structure stays the same (with buildtool separate from the monorepo)?

Curious what @karlskewes has to say about this.

rfc/monorepo.md
jcavanagh: Definitely makes sense to include it until it's officially replaced - I'll add it here as well.

jcavanagh:
@dbyron-sf Thank you!  Appreciate the review!

> It doesn't build with gradle, and I haven't thought this all the way through, but could it make sense to include [spinnaker/buildtool](https://github.com/spinnaker/buildtool) in the monorepo? Perhaps it makes sense to include some detail [here](https://spinnaker.io/docs/releases/release-manager-runbook/) about release mechanics with a monorepo. The current process is documented here, but it basically involves tagging the HEAD of the relevant release branch in each repo and then running a [buildtool github action](https://github.com/spinnaker/buildtool/blob/f7c125019e4b7e67e9bf7f182ae4963fc854fdf9/.github/workflows/release.yml).

I think this definitely makes sense - the monorepo must have everything it needs to build, test, and publish.  `buildtool` will likely have a much easier time of things with all the code it needs right there.  Many of our tools that did similar things ended up in a much simpler, streamlined form post-monorepo, and it was certainly a positive thing.  

> Maybe the release process with a monorepo will still involve pushing a tag to release artifacts (i.e. docker containers, debian packages, jars)? And maybe we'll still use buildtool to publish boms, etc.? So though buildtool will need some adjustment, maybe the general structure stays the same (with buildtool separate from the monorepo)?

The conceptualization of the tags are a little different, since a tag maps to a Git commit, and a single Git commit can cut across what would have previously been many individual service tags.  However, tags do still make sense in the context of the monorepo as a whole.

rfc/monorepo.md
jcavanagh: > I think we should NOT include spinnaker-monitoring It's not even referenced on the docs as a supported system anymore. Just hasn't been removed from halyard... and probably should figure out how to make sure it's defaulted to observability plugin instead. BUT kinda been waiting to see where "how-to-deploy-spinnaker" stuff goes

I do agree that `spinnaker-monitoring` is de-facto replaced for most major users, but we can always delete it from the monorepo once it's actually replaced.  I think it is likely to be more total effort to exclude it, as that would complicate the build and release process until it is removed.

dbyron-sf:
> The conceptualization of the tags are a little different, since a tag maps to a Git commit, and a single Git commit can cut across what would have previously been many individual service tags. However, tags do still make sense in the context of the monorepo as a whole.

Agreed.  I figure we still need to do something to indicate that a particular commit is e.g.  2023.0.

rfc/monorepo.md
karlskewes: Yes, `spinnaker-monitoring` repo is frozen.
The version 1.4.0 is pinned in Halyard's BOM.
During a release we re-tag the existing container image with the friendly `spinnaker-1.x.y` so it's easy to consume.

We can continue to do something similar in monorepo CI until support for it is removed.
We don't need to and shouldn't include it in a monorepo.

jcavanagh: If it isn't built as part of a release, then I agree it should be excluded.  I'll revise to that effect.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+
+Currently, each Java project within Spinnaker maintains its own semver-shaped version designation pinned within various other Spinnaker projects.  This is unnecessary with composite builds, as the project builds as one cohesive unit without any dependency versions at all.
+
+However, containers, libraries, and other artifacts must still be published, and there is a need for a contract with library consumers (such as plugins).
+
+#### Spinnaker Release Versioning and Scope
+
+Spinnaker releases will use a date-based versioning scheme of `<year>.<number>.<patch>`, where `number` represents the number of major releases in that calendar year, and `patch` designates new builds produced for reasons outlined below.  The first numbered release of 2023 would therefore be `2023.0`, followed by `2023.1` and so on.  Patches to numbered releases are appended to the numbered release, e.g. `2023.0.1`.
+
+The initial version of a numbered release is considered to have patch version zero, even if not explicitly stated for brevity.  Unqualified numbered releases refer to the latest patch of that numbered release.
```
karlskewes, line 167: I think for strict semver systems like Jar's and Debian packages we will need to always publish artifacts tagged as `{major}.{minor}.{patch}`? i.e: always add `.0` patch on the end.

We could re-tag a `<year>.<number>` container image when a new patch comes out if there is appetite for consuming those?
This would look like container images such as alpine where the same container SHA might be tagged: `3.17.1`, `3.17`, `3`, `latest`

karlskewes: > Unqualified numbered releases refer to the latest patch of that numbered release.
I don't think we can re-publish jar's and debs so it's only containers where we have the luxury of tags.

Apologies if I have misread.

karlskewes:
> > It doesn't build with gradle, and I haven't thought this all the way through, but could it make sense to include [spinnaker/buildtool](https://github.com/spinnaker/buildtool) in the monorepo? Perhaps it makes sense to include some detail [here](https://spinnaker.io/docs/releases/release-manager-runbook/) about release mechanics with a monorepo. The current process is documented here, but it basically involves tagging the HEAD of the relevant release branch in each repo and then running a [buildtool github action](https://github.com/spinnaker/buildtool/blob/f7c125019e4b7e67e9bf7f182ae4963fc854fdf9/.github/workflows/release.yml).
> 
> I think this definitely makes sense - the monorepo must have everything it needs to build, test, and publish. `buildtool` will likely have a much easier time of things with all the code it needs right there. Many of our tools that did similar things ended up in a much simpler, streamlined form post-monorepo, and it was certainly a positive thing.
> 
> > Maybe the release process with a monorepo will still involve pushing a tag to release artifacts (i.e. docker containers, debian packages, jars)? And maybe we'll still use buildtool to publish boms, etc.? So though buildtool will need some adjustment, maybe the general structure stays the same (with buildtool separate from the monorepo)?
> 
> The conceptualization of the tags are a little different, since a tag maps to a Git commit, and a single Git commit can cut across what would have previously been many individual service tags. However, tags do still make sense in the context of the monorepo as a whole.

All sounds great. 

We could modify `buildtool` to do monorepo releases.
There are existing OSS GHA's that can be leveraged for parts of the release process but buildtool has it all already. Can iterate later.

A lot of the buildtool code base is designed for multiple git repositories or unused (integration tests with archived dependencies).
Instead of importing `buildtool` I'd be tempted to leave it separate and PR a simplified version to the monorepo.

rfc/monorepo.md
jcavanagh: > I think for strict semver systems like Jar's and Debian packages we will need to always publish artifacts tagged as {major}.{minor}.{patch}? i.e: always add .0 patch on the end.

I'm not sure about Debian packages, but jars aren't that strict about their versioning - we mostly just need to be conscious of what Gradle considers the "ordering" to be.  

> I don't think we can re-publish jar's and debs so it's only containers where we have the luxury of tags.

That is true - that line was only meant to refer to images and usage in the general vernacular.  I'll clarify it.  

> We could re-tag a <year>.<number> container image when a new patch comes out if there is appetite for consuming those?

I'm sure there would be - we make a lot of use of `<release>.x`-tagged containers, and I would guess that many others do too.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+
+1. Merge changes from the individual repo to the monorepo using the `subtree` strategy
+    - `git merge -X subtree=<subtree> <remote_name>/<branch>`
+    - Repeat for each remote project and each branch that needs reintegration
+
+This is a simple shell script that implements a process like the above:
+
+<details>
+<summary>Expand pull.sh</summary>
+
```
jasonmcintosh, line 291: Wondering if this should be a git check on merge kinda thing to make sure things are updated?  Just ... would hate to have it pure manual kinda thing.  Maybe a gradle task sorta like the spotless rules to verify subtrees are there/updated/etc.

jasonmcintosh (approved):
Overall good direction i like it a few remaining questons:
* How does this change/handle backports would be good to call out.  E.g. "With a mono repo, backports to a non-monorepo still work on the monorepo via XYZ" kinda thing or better explain that.
* We have halyard merged in. Kayenta is a challenge here for a few reasons.  Kayenta CURRENTLY can operate entirely as a standalone service.  That ALSO impacts some of the UI aspects of kayenta.  A number of organizations use/run kayenta without spinnaker and this really should continue to be an option.  I don't see a monorepo conflicting with this paradigm, but this is ONE place I'd like to see addressed more directly
* From a security/governance perspective.  There were a few governance groups that likely need cleaning up - e.g. I think there were spinnaker groups for "clouddriver only" or "kayenta" only, and that may change as a result of a single vs. multiple repos.
* Only other thought might be if there are any impacts to docs system..  E.g. updating docs/contributor docs to include all this information would be GREAT!  

I'm definitely in favor as I think this'll help a LOT on multiple levels!  AND... I could see this for example LONG term negating the need for the BOM or other complexities of how deploys operate today!  Going ahead adding "approved" as I think this is what we should move to :)

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+The initial version of a numbered release is considered to have patch version zero, even if not explicitly stated for brevity.  Unqualified numbered releases refer to the latest patch of that numbered release.
+
+Significant new feature work should result in a new release number.  Bugfixes, security fixes, or other important non-breaking, non-feature changes may be introduced in patch releases.  Small updates to existing functionality may be considered for patch releases if the changes are additive and non-breaking - this is left to the judgement of the reviewer and approver, along with the release manager.
+
+All patch releases (`<year>.<number>.<patch>`) should be fully compatible with that version's initial release, and all previous patches to that release number.  Patch releases should be fully non-breaking, from external API all the way down to public function signatures in libraries.  Exceptions can be made to mitigate critical bugs or security issues.
```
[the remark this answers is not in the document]
jcavanagh: I don't think it should be an issue - `2023.0.1` is perfectly valid semver, it just has a bigger-than-usual major version.

rfc/monorepo.md
jcavanagh: We could, but it's a purely transitional process - once active development has moved to the monorepo, it won't be needed anymore.  

It's also possible that manual conflict resolution could be required if there is new development in the monorepo not present in the individual projects.

My suggestion would probably be a GHA manual or scheduled trigger that runs and auto-commits successful merges, with the possibility of a human running it also available.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+
+Linking the builds of `deck`, and `deck-kayenta` is not something Gradle can do, but we have two good options here:
+
+1. Use the local `deck-kayenta` code during the build process of `deck`, using `yarn`'s workspaces feature
+1. Just integrate `deck-kayenta` as a regular `deck` module, like all the other Lerna subprojects under `packages/*`
+
+### The `spin` CLI
+
+This project is actually the most difficult to integrate into the monorepo, as golang builds tend to be really finicky about their environment.  I think Github Actions should afford us the flexibility to partition that out reasonably, but that requires some investigative work.
+
```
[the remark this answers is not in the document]
jcavanagh: I am not familiar with `goreleaser`, but definitely open to it.  GHA can use a separate container per step, so I don't actually anticipate too much trouble porting the existing process to a monorepo build, but thought it was worth a callout.

jcavanagh:
@jasonmcintosh Thank you!  Answered your questions below:

> How does this change/handle backports would be good to call out. E.g. "With a mono repo, backports to a non-monorepo still work on the monorepo via XYZ" kinda thing or better explain that.

There are sections in the document about how Git changes can be moved from one type of repository to another.  The [section relevant to this question](https://github.com/spinnaker/governance/pull/336/files#diff-a56c8a7010c73454be672ac2b81d8df1e145318115a9a833fa9570da4b1dc36cR376) describes how to do this - the process of backporting a change ad-hoc to an OSS single-project repo is identical to the process that a third-party private single-project fork would use.  

It is worth noting that the backport process from monorepo -> single-project repos must be performed N times, where N is the number of services the change to be backported cuts across.  Each partial diff would apply the changes relevant to that single-project repository.

> We have halyard merged in. Kayenta is a challenge here for a few reasons. Kayenta CURRENTLY can operate entirely as a standalone service. That ALSO impacts some of the UI aspects of kayenta. A number of organizations use/run kayenta without spinnaker and this really should continue to be an option. I don't see a monorepo conflicting with this paradigm, but this is ONE place I'd like to see addressed more directly

I have a section about `deck-kayenta` [here](https://github.com/spinnaker/governance/pull/336/files#diff-a56c8a7010c73454be672ac2b81d8df1e145318115a9a833fa9570da4b1dc36cR132), which could use some opinions.  What we do today is link the builds so that `deck` can basically do the equivalent of a composite build with `deck-kayenta`.  `deck-kayenta` will still have its own published artifacts, even if `deck` builds with it directly.  

Kayenta the service will still be available in its current containerized and library forms.  It's standalone-ness from that point is entirely configuration-based as far as I know.  

I'll add a section about standalone Kayenta to clarify these points.

> From a security/governance perspective. There were a few governance groups that likely need cleaning up - e.g. I think there were spinnaker groups for "clouddriver only" or "kayenta" only, and that may change as a result of a single vs. multiple repos.

Good callout - I'm not very familiar with those, so I'd leave that to the TOC (or whoever is most suited for that).

> Only other thought might be if there are any impacts to docs system.. E.g. updating docs/contributor docs to include all this information would be GREAT!

Definitely plenty of developer documentation changes needed, and also changes to Halyard's features regarding checking out the projects, its assumption of `~/dev` for most of those paths, and so on.  

This would be a great opportunity to get Halyard out of the Getting Started docs, and just replacing it with Gradle (or something else).  It's not needed to deploy Spinnaker to a k8s cluster to run it, and all that would be needed is a stock set of default configuration that would typically be generated by Halyard.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+Gradle composite can also build plugins to be used in the same build, allowing us to also combine [Spinnaker's Gradle build plugin](https://github.com/spinnaker/spinnaker-gradle-project) into the project itself.
+
+This is the only non-Spinnaker-service version that was pinned in Spinnaker Java projects, and including `spinnaker-gradle-project` into the monorepo enables trivial iteration on the Gradle build process.
+
+### Releases and Versioning
+
+Currently, each Java project within Spinnaker maintains its own semver-shaped version designation pinned within various other Spinnaker projects.  This is unnecessary with composite builds, as the project builds as one cohesive unit without any dependency versions at all.
+
+However, containers, libraries, and other artifacts must still be published, and there is a need for a contract with library consumers (such as plugins).
```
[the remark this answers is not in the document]
jcavanagh: >we'll still publish individual containers for clouddriver, orca, etc. OR is the intent to have a SINGLE docker with ALL files?

Yes, individual containers just like today.  I don't think there's a use case for a universal container.

> JARs will be published for cloudriver on every merge of resources like orca or fiat due to dependencies

We don't actually need to publish jars outside of numbered/patch releases (or whatever latest/nightly cadence from `main` we choose to implement).  Those are purely for external library consumers.  Gradle builds will produce jars directly from the code in the repository to be shipped with containers, for example - no external Maven round-trip required.  

For `main` library publishing, I would very much recommend that jars get a full publish cycle each time it is done.  This does create some identical published artifacts with different versions, but there are some weird issues with generation of Maven POMs via the `maven-publish` plugin and [gradle module metadata](https://docs.gradle.org/current/userguide/publishing_gradle_module_metadata.html) that are difficult to work around.  

Both of those things will write specific dependency versions into their respective manifests shipped up to a Maven repository.  The problem is that the "version" in there is entirely made-up - it's whatever we want it to be.  Gradle will build everything according to the "version" that CI says it should be, and that will the version written into POMs and gradle module metadata.  If those jars don't actually exist post-publish, pulling any of the platform modules (all Java `-bom` packages e.g. `kork-bom`) will fail because they won't find the dependencies that they think should exist.  

The alternative is a bunch of trying to find the "last" version of every dependency and somehow telling Gradle about that - an approach I would strongly recommend against.  It is far simpler to just publish every jar with every release.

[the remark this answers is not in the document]
jcavanagh: You can see an example of what a platform Maven BOM looks like here (latest `orca-bom`):
https://search.maven.org/artifact/io.spinnaker.orca/orca-bom/8.27.1/pom

Within that you can see specific versions for `fiat`, `kork`, and `orca` libraries, all of which must exist or builds against that `orca-bom` version will fail.

rfc/monorepo.md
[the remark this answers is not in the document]
karlskewes: [GoReleaser](goreleaser.com) is great. 
For the Java services we have our custom gradle publishing plugin.
For `spin` we manually wrote the artifact publishing steps in our GHA instead of taking a dependency on GoReleaser. We could change but I'm not sure what we would get.

When Gate API's are changed we need to generate swagger def and then generate `spin` code from the swagger def.
The steps are documented [here](https://github.com/spinnaker/spin/blob/master/CONTRIBUTING.md#updating-the-gate-api).
We can run a GHA to check that swagger is up to date and spin code is the same, potentially fail the build if not.
Currently we manually refresh the code from swagger when cutting a release - see [step 12 of release runbook](https://spinnaker.io/docs/releases/release-manager-runbook/#one-week-after-branches-are-cut).

Overall we can use all of our existing `spin` GHA configuration with minor changes to repo paths and similar. It will run on its own VM.

rfc/monorepo.md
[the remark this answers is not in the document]
jasonmcintosh: SO a key thing to publish still jar wise is plugins.  If you're developing a plugin that uses say clouddriver-core, you need it as a test compile only dep in order to build.  NOT having JARs published would make plugin dev more difficult I'd think.

rfc/monorepo.md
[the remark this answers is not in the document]
jasonmcintosh: THat's the point - it's valid semver, but halyard doesn't obey standard semver.  Will try to find the code on that particular issue.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+
+Release branches will have a naming scheme similar to today, but date-oriented.  A release branch is prefixed by `release-`, appends the release version, and is suffixed with `.x`, e.g. `release-2020.3.x`.
+
+This branch will produce containers, libraries, and other artifacts for that numbered release.
+
+#### Artifact Versioning Scheme
+
+Spinnaker library and container versions from release branches will mirror the release version (e.g. `2023.3.4`).  This should be defined in the root `gradle.properties` file, and used throughout included builds and subprojects to publish their respective artifacts.
+
+From `main`, library and container versions will be published as `main-<gha_run_id>`, e.g. `main-4041166540`.  Github Actions run IDs always increase, so this functions identically to timestamp-based snapshot versions with two key benefits:
```
jasonmcintosh, line 193: I could see doing commit hash to allow tracking jars to a specific commit.  GHA Run id isn't bad and does allow easy view of order.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+    - A Docker container named `main-4041166540` or jar named `clouddriver-aws-main-4041166540` can be found at `https://github.com/spinnaker/spinnaker/actions/runs/4041166540`
+1. The shorter build ID is a lot more human readable, and there is no need for.
+
+This number follows [Gradle's rules for version ordering](https://docs.gradle.org/current/userguide/single_versions.html#version_ordering) to allow prefix patterns and version ranges to work naturally.  For example, the latest build of the `main` branch's Java libraries can be used by asking Gradle to find version `main-+`.
+
+Individual services/libraries will no longer have semver-shaped versions of their own.  Since all code is always built and published together, there is no need to declare any inter-service compatibility.  The compatibility contract is instead made against the project as a whole, rather than individual components.
+
+Examples:
+`main` branch publishes as: `<package>-main-<build_number>.jar`, e.g. `clouddriver-kubernetes-main-4041166540.jar`
+`release-<version>` branches publish as: `<java_package>-release-<full_version>.jar`, e.g. `clouddriver-kubernetes-release-2023.3.4.jar`
```
jasonmcintosh, line 205: MAIN concern with this is backports.  Given the above... is the plan for backports to go to "release-2022.3.6" from a "release-2022.3.5" version?  OR is it a "release-2022.4.0" situation?  When is a backport a minor vs. patch release?

rfc/monorepo.md
jcavanagh: GHA run ID does directly map to a commit, and we can include that in jar metadata if we want the sha directly.  I don't usually prefer having a sha in the jar name, since it makes the library more difficult to consume.  Gradle ordering with a sha can be very non-obvious, and if we happen to get an all-numeric short sha, it gets weirder still.

rfc/monorepo.md
jcavanagh: A backport would only be to a different numbered release (the second number, where the semver "minor" slot would be).  

> is the plan for backports to go to "release-2022.3.6" from a "release-2022.3.5" version?

`release-2022.3.6` would fully replace version `release-2022.3.5`, due to the non-breaking contract of a patch change described in the proposal.  There is no separate branch for these two versions - that scenario would be equivalent to `1.29.1` vs `1.29.2` today.

There is no concept of a "minor" release in this proposal - the suggested format is no contract between numbered releases (where the "minor" slot is), since that's a fuzzy judgement call that never seems to work out all that well in my experience.

rfc/monorepo.md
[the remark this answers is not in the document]
jasonmcintosh: https://github.com/spinnaker/halyard/blob/a036362b27581abeb83de99d682f0e200d12f1e1/halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/Versions.java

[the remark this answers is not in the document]
jcavanagh: Looks like it would be alright to me - it just does comparisons on each part, and numbers are comparable.  We can add some tests to be sure.

rfc/monorepo.md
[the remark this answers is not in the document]
jcavanagh: I agree, which is why I recommended a full publish for all jars on release branches and whatever cadence for `main` we choose.  Publishing for plugin development is explicitly called out in the proposal.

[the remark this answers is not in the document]
edgarulg: Publish all jars for release branches and main will help Spinnaker plugin developers with their CI process. Glad to see it as plan.

rfc/monorepo.md
```diff
@@ -0,0 +1,557 @@
...
+
+This allows much simpler consumption of Spinnaker libraries, particularly in the case of plugins.  In this publishing form, building and testing plugins against many different Spinnaker releases can be achieved entirely via [Gradle's dependency resolution configuration](https://docs.gradle.org/current/userguide/resolution_rules.html), or via Gradle properties.  This is an extremely powerful feature that is critical to our everyday workflows.
+
+### Plugin Repository Compatibility
+
+The plugin system has a feature to allow a plugin repository entry to declare compatibility with certain Spinnaker service versions.  Those values are currently the service-specific semver-shaped library versions.
+
+That compatibility declaration may need to be reworked to support date-oriented release numbering.  Code within the Spinnaker plugin system implementation may need to be reworked to accommodate the new Spinnaker library versioning.
```
edgarulg, line 213: I like it. We need more flexibility in version schema.

jcavanagh:
@link108 @dbyron-sf I am about midway through the initial POC implementation.  Is there any additional feedback you'd like to give in the meantime?

dbyron-sf:
> @link108 @dbyron-sf I am about midway through the initial POC implementation. Is there any additional feedback you'd like to give in the meantime?

Only best of luck and thanks so much for doing it!  Shall we merge the PR?

dbyron-sf (approved):

mattgogerly (approved):

link108 (approved):

link108:
woot woot! This looks good, if there are no objections, I'm going to merge this tomorrow, 3/17

karlskewes (approved):

edgarulg (approved):

JonGilmore:
>  is Keel in scope to be included in the monorepo? It's quite the ugly duckling at the moment - I don't know anybody outside of Netflix (and maybe AirBnB?) using it. There's been a lot of issues getting Kork changes to merge.

@mattgogerly just wanted to chime in here, Samsung SmartThings is heavily relying on keel and have ~50 production services deployed using it at the moment. Our intention is to push the entirety of our services to MD in the future.

PR #341, opened 2023-03-23
update: finish up TODOs for cloud SIG

xibz:


sig-cloud/README.md
```diff
@@ -56,9 +56,9 @@ _OpsMx_
 
 
 ### Meeting Cadence
- - 1 hour, monthly, on **[Zoom](TODO-need-zoom-link)**
+ - 1 hour, monthly, on **[Google Meet](https://8x8.vc/greencapsdesirepolitely)**
```
dbyron-sf, line 59: This doesn't look like a google meet link, but that probably doesn't matter.

dbyron-sf (approved):

sig-cloud/README.md
link108:
```suggestion
 - 1 hour, monthly, on **[Google Meet](https://meet.google.com/afb-rppx-zye)**
```

link108 (approved):
left a suggestion, but feel free to merge when ready!

sig-cloud/README.md
xibz: Ah yea, I think Google Meet uses 8x8 as the backend now, so the link's are basically the same thing, with different frontends.

PR #345, opened 2023-06-29
rfc(tracing): Adding Distributed Tracing to Spinnaker

mattgogerly:
Let's make troubleshooting easier.


rfc/distributed-tracing.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+The next step will be to choose between [OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/traces/) and
+[Brave](https://zipkin.io/pages/tracers_instrumentation.html). Brave was added to Spinnaker in
+[kork#744](https://github.com/spinnaker/kork/pull/774) and is already wired up to OkHttp.
+
+Both [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/okhttp/okhttp-3.0/library)
+and [Zipkin](https://github.com/openzipkin/brave/blob/master/instrumentation/okhttp3) provide instrumentation for OkHttp,
+therefore we are free to choose. OpenTelemetry has been gaining a lot of momentum, but Brave is also widely used.
+I leave this choice for discussion in this RFC.
```
dbyron-sf, line 80: I'm beginning to get educated here and am leaning to OpenTelemetry.  Need to get more educated to form a proper preference.

dbyron-sf:
This is exciting!  Thank you @mattgogerly !

mattgogerly:
It turns out Micrometer Tracing doesn't support SpringBoot < 3.0, so we'll have to use Sleuth until we get there.

I'll update the RFC accordingly.

mattgogerly:
I've been playing around a bit and the proposed approach seems to work. Need to do a bit more testing to prove it out further.

Need somebody with more knowledge of Gradle to confirm if we can include the Sleuth dependencies in kork-runtime rather than adding them as implementation dependencies  to every service explicitly.

PR #350, opened 2024-03-14
RFC: pipeline-builder library

nicolasff:
This RFC proposes to add the `pipeline-builder` Java library to the set of Spinnaker repositories. `pipeline-builder` helps developers write and maintain pipelines by writing fluent, readable, and testable code as an alternative to the internal JSON format used by Spinnaker.

---

#### Links

- Repository: https://github.com/nicolasff/pipeline-builder
- Tutorial: [examples/README.md](https://github.com/nicolasff/pipeline-builder/blob/main/examples/README.md)
- Examples: [7 classes, each building a pipeline](https://github.com/nicolasff/pipeline-builder/tree/main/examples/src/main/java/com/apple/spinnaker/examples)

### DIFF
--- a/rfc/pipeline-builder.md
+++ b/rfc/pipeline-builder.md
@@ -1,0 +1,234 @@
+# Pipeline Builder Library
+
+|               |                                                                    |
+| ------------- |--------------------------------------------------------------------|
+| **Status**    | Proposed                                                           |
+| **RFC #**     | [350](https://github.com/spinnaker/governance/pull/350)            |
+| **Author(s)** | Nicolas Favre-Felix ([`@nicolasff`](https://github.com/nicolasff)) |
+| **SIG / WG**  | Platform SIG                                                       |
+
+## Overview
+
+Add a new Java library to Spinnaker that provides a clean API with a fluent/builder style to build and maintain pipelines as testable code instead of storing and editing them in JSON directly.
+
+Repository: https://github.com/nicolasff/pipeline-builder
+
+### Example
+
+A pipeline is built by creating a class that extends the `JsonPipelineBuilder` abstract class and implements two methods:
+
+- `getUniqueName()`: returns a unique name for the pipeline, used to generate a stable pipeline definition ID.
+- `buildPipeline()`: returns a `Pipeline` object, which can then be serialized to `String` with `toJson()`.
+
+<details>
+<summary>Expand to see a complete pipeline builder class</summary>
+
+```java
+public class TutorialPipelineBuilder extends JsonPipelineBuilder {
+
+    @Override
+    public String getUniqueName() {
+        return "examples.wait-and-evaluate";
+    }
+
+    @Override
+    protected Pipeline buildPipeline() {
+        Stage waitStage = Stage.builder()
+            .type(StageTypes.WAIT)
+            .name("Wait a moment")
+            .contextObject(WaitContext.ofSeconds(5))
+            .build();
+
+        Stage evalSumStage = Stage.builder()
+            .type(StageTypes.EVALUATE_VARIABLES)
+            .name("Evaluate sum")
+            .parentStage(waitStage)
+            .context(Map.of("variables", List.of(
+                Map.of(
+                    "key", "sum",
+                    "value", "${ #toInt(parameters.a) + #toInt(parameters.b) }"))))
+            .build();
+
+        return Pipeline.builder()
+            .parameters(List.of(
+                PipelineParameter.builder()
+                    .name("a")
+                    .defaultValue("17")
+                    .build(),
+                PipelineParameter.builder()
+                    .name("b")
+                    .defaultValue("25")
+                    .build()))
+            .name("Tutorial")
+            .stages(List.of(waitStage, evalSumStage))
+            .build();
+    }
+}
+```
+
+</details>
+
+Once built as a `Pipeline` object, the corresponding JSON can be generated easily:
+```java
+    Pipeline pipeline = new TutorialPipelineBuilder().build();
+    String json = pipeline.toJson();
+```
+
+### Goals and Non-Goals
+
+Goals:
+
+- Provide a clean, fluent API to build and maintain Spinnaker pipelines as code
+- Leverage Java features and tools to make pipelines more readable and maintainable, such as:
+  - Version control
+  - Type safety
+  - Testability
+  - IDE support
+  - Code completion
+  - Refactoring
+- Focus on the developer experience, and offer simpler alternatives when the underlying structure is complex or obscure
+- Let users validate their pipelines with code, adding their own tests to the library's built-in safeguards
+
+Non-goals:
+
+- Replace the Deck web interface
+- Replace the JSON format
+- Upload pipelines to Spinnaker or access the Spinnaker API in any way
+- Convert pipelines from JSON to Java
+
+## Motivation and Rationale
+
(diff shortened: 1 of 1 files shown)


rfc/pipeline-builder.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Alternatives and Prior Art
+
+Some users already build and maintain their pipelines using other generators, often based on a different source language or on a templating language. Examples include [HCL](https://github.com/hashicorp/hcl) and [Jsonnet](https://jsonnet.org/) with [Sponnet](https://github.com/spinnaker/sponnet).
+
+While these tools help simplify the syntax of the pipeline definitions, they do not provide the same level of type safety, validation (with tests), refactoring support, and IDE compatibility as a library written in the same language as most of the Spinnaker codebase. They usually have some of the same drawbacks as the JSON format, in particular when it comes to poorly-named fields or the complex structures behind some common stage options.
+
+They are also not generally testable with code, unless these tests are limited to validating the generated JSON.
+
+## Evolution with Spinnaker
```
dbyron-sf, line 217: Let me start by saying I'm a fan of this effort.  All by itself I expect it would be quite useful.  I'm curious if you/Apple have concrete plans to make the kinds of changes to orca that this section discusses.  Getting a "real" connection to spinnaker seems key to keeping it from getting stale.

I'm also curious if you considered an approach with jsonschema to define the structure of pipelines/various stages and then automated code generation from there.  I'm thinking about facilitating validation, both inside spinnaker itself, but also externally.  It's arguably a separate concern, but they're related.

nicolasff: > I'm curious if you/Apple have concrete plans to make the kinds of changes to orca that this section discusses. Getting a "real" connection to spinnaker seems key to keeping it from getting stale.

The topic of interface-only modules is something we've talked about in the Spinnaker team, but more in the sense that we were lamenting the lack of a clear separation of interfaces and implementations (and wishing for this structure to exist) than as an actual plan to take this on. That said, I see two points to take into consideration here:

1: Extracting interfaces for the context classes could be done gradually, maybe adding them in batches for related stages for examples (one PR with all the Kubernetes types, the next one for stages related to artifacts, etc.)

2: The [existing context classes](https://github.com/nicolasff/pipeline-builder/tree/main/pipeline-builder/src/main/java/io/spinnaker/pipelinebuilder/json/contexts) that we already include do not always provide the exact same names for fields and methods as the names used in the actual `StageData` context class from Orca. This is because a major focus of this library is developer experience, and the representation used internally isn't always easy to understand. The existence of two "representations" for some fields – Java name vs JSON name – is something we would have to consider, just like we would have to ask whether it makes sense to have builder annotations on some classes and fields that we publish from an `orca-stages-api` module.

For example, take the (very basic) [context class](https://github.com/nicolasff/pipeline-builder/blob/9bf07ff40f3973d10c3be43b600e4e09054d3af8/pipeline-builder/src/main/java/io/spinnaker/pipelinebuilder/json/contexts/WaitContext.java#L40-L44) we created for the Wait stage:

```java
@Getter
@RequiredArgsConstructor(staticName="ofSeconds")
public class WaitContext implements ContextObject {
    @JsonProperty("waitTime") private final int waitTimeSeconds;
}
```

[This pipeline](https://github.com/nicolasff/pipeline-builder/blob/9bf07ff40f3973d10c3be43b600e4e09054d3af8/examples/src/main/java/com/apple/spinnaker/examples/ContextObjectsPipelineBuilder.java#L22-L26) shows how it is used with a Stage builder instead of passing a `Map` with `String` keys:

```java
        Stage waitStage = Stage.builder()
            .type(StageTypes.WAIT)
            .name("Wait a moment")
            .contextObject(WaitContext.ofSeconds(5))
            .build();
```
For reference, the approach without a dedicated class is to call `.context()` instead of `.contextObject`, [like this](https://github.com/nicolasff/pipeline-builder/blob/9bf07ff40f3973d10c3be43b600e4e09054d3af8/examples/src/main/java/com/apple/spinnaker/examples/TutorialPipelineBuilder.java#L23):

```java
            .context(Map.of("waitTime", 5))
```

This `WaitContext` class would be usable as a context class in Orca (maybe with some setters for Jackson), but as a consumer of `orca-stages-api` it would look a bit weird to have this static `ofSeconds` constructor in there. This is a simple example, but we could certainly have some that are significantly more complex.

---

> I'm also curious if you considered an approach with jsonschema to define the structure of pipelines/various stages and then automated code generation from there.

Not with jsonschema, no.

But some of the teams that use our Spinnaker instance manage their pipelines using [Pkl](https://pkl.apple.com/), a recently-released "configuration-as-code" language. They wrote a Pkl library with pre-defined types for common Spinnaker data structures, so users of this library build and maintain their pipelines in a type-safe, VCS-compatible, reviewable, and testable language.

These two solutions emerged independently to address the same problem; I thinks it says something about the usability of JSON, or at least our perception of JSON as a usable format for this purpose.

dbyron-sf: All of this sounds great.  I'm a little nervous to use a new tool like Pkl, but the various attempts to use jsonschema over the years have failed and Spinnaker is sorely lacking validation like this.  So overall, I'd say yes please to all of it.

Extracting interfaces feels like it's big enough to be its own project.  Would you be willing to extract the first one, or perhaps write up some details on what to do / how to do it and then other folks could chip in to move that forward?

dbyron-sf (approved):

dbyron-sf:
We talked at the TOC meeting this morning around the mechanics of this.  We could make a separate repo in the spinnaker github org for this, but we were leaning towards waiting for the monorepo.  @jcavanagh or @nicolasff curious on your thoughts here.

dbyron-sf:
> @dbyron-sf Thanks for the review and feedback!
> 
> I talked to @jcavanagh today about the monorepo structure since he's currently working on it, and he agreed with the suggestion to move this repo to the `spinnaker` org before referencing it [using a git subtree](https://github.com/spinnaker/governance/blob/d21c7b66c4ae9e90ddcfaf39207099a634972b5a/rfc/monorepo.md#monorepo-creation-and-fork-management) in the upcoming monorepo.
> 
> We already track the `pipeline-builder` repo in our own monorepo, so Joe will be able to use a similar structure in his current work to integrate this library after its repo is moved to the `spinnaker` org.
> 
> David, let me know when I can transfer ownership. I've looked at the settings in my repo and there doesn't seem to be anything preventing me from transferring it to the `spinnaker` org. The [GitHub docs](https://docs.github.com/en/enterprise-cloud@latest/repositories/creating-and-managing-repositories/transferring-a-repository) explain that an email will be sent to the new owner, who needs to approve it within 24 hours or the transfer request will expire. It's ready now, having just been updated with [a few commits](https://github.com/nicolasff/pipeline-builder/commits/main/?since=2024-01-10&until=2024-04-08) we had made since it was published.

It doesn't sound like we need to do anything in the spinnaker org to prepare for this (e.g. creating a repo), so I say please go ahead and transfer it.  Thanks much!

nicolasff:
@dbyron-sf I am not allowed to transfer the repo directly to the org. Should I try sending it to you instead?

image

dbyron-sf:
Yeah, if there's some other way to initiate the transfer from your side, please do give it a whirl.

nicolasff:
Thanks, sent!

image

dbyron-sf:
Excellent, and I moved it again, so now it's https://github.com/spinnaker/pipeline-builder.

